PR #112, opened 2023-08-11
COMPENF-126-Display marker on map based on coordinates

barrfalk:
# Description

Adds a marker to a map based on the complaints location.

# How Has This Been Tested?

- [x] View the details of a human wildlife complaint, ensure that the marker on the map corresponds with the lat/long coordinates when present.
- [x] View the details of a human wildlife complaint, ensure that the map does not display.  Note that upcoming functionality will ensure that the map will always display when we're able to reverse geocode the community to a lat/long coordinate.
- [x] View the details of a enforcement complaint, ensure that the marker on the map corresponds with the lat/long coordinates when present.
- [x] View the details of a enforcement complaint, ensure that the map does not display.  Note that upcoming functionality will ensure that the map will always display when we're able to reverse geocode the community to a lat/long coordinate.

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-112-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-112-frontend.apps.silver.devops.gov.bc.ca/) available


marqueone-ps (changes requested):
@barrfalk just a couple changes need to be made before we're ok to merge into main.

- you'll want to rename LeafletMapWithPoint.tsx to use kebab-case 
- not critical, and can wait, but it would be good to do now if we can, use font-awesome for the leaflet icons instead of the images that you're currently using
- for reusability of the map ComplaintLocation I would recommend decoupling the complaint from the component and instead pass the latitude and longitude as props, this way if we need to use a map for something else we shouldn't have to create a new component.

marqueone-ps (approved):

PR #269, opened 2024-01-29
fix: schema-spy

mishraomp:
1. Removed deprecated GHA(https://github.com/joshuaavalon/flyway-action)
2. Fix the error during workflow run https://github.com/bcgov/nr-compliance-enforcement/actions/runs/7674137896/job/20918154521#step:7:66

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-269-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-269-frontend.apps.silver.devops.gov.bc.ca/) available

### DIFF
--- a/.github/workflows/schemaspy.yaml
+++ b/.github/workflows/schemaspy.yaml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-22.04
     services:
       postgres:
-        image: postgis/postgis
+        image: postgis/postgis:16-3.4
         env:
           POSTGRES_DB: default
           POSTGRES_USER: postgres
@@ -28,16 +28,14 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - uses: joshuaavalon/flyway-action@v3.0.0
-        name: Generate SchemaSpy docs for node backend
-        with:
-          url: jdbc:postgresql://postgres:5432/postgres
-          user: postgres
-          password: default
+      - uses: docker://flyway/flyway:10
         env:
-          FLYWAY_VALIDATE_MIGRATION_NAMING: true
+          FLYWAY_URL: jdbc:postgresql://postgres:5432/default
+          FLYWAY_USER: postgres
+          FLYWAY_PASSWORD: default
           FLYWAY_LOCATIONS: filesystem:./backend/db/migrations
           FLYWAY_DEFAULT_SCHEMA: "public"
+
       - name: Create Output Folder
         run: |
           mkdir output
--- a/.github/workflows/schemaspy.yaml
+++ b/.github/workflows/schemaspy.yaml
@@ -28,7 +28,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v4
-      - name: setup Flyway
+      - name: Flyway
         uses: docker://flyway/flyway:10
         env:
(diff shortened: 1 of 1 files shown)


mishraomp:
that did not work :( @barrfalk @afwilcox , need to see why, I will mark this PR as draft
Gathering schema details......ERROR - Failed to retrieve comment for catalog 'default' using SQL 'SELECT datname AS catalog_name, pg_catalog.obj_description(datlastsysoid, 'pg_database') AS catalog_comment FROM pg_catalog.pg_database WHERE datname = :catalog ;'
ERROR: column "datlastsysoid" does not exist
  Position: 60
WARN  - Failed to retrieve stored procedure/function details using sql 'select r.routine_name || '(' || oidvectortypes(p.proargtypes) || ')' as routine_name, case when p.proisagg then 'AGGREGATE' else 'FUNCTION' end as routine_type, case when p.proretset then 'SETOF ' else '' end || case when r.data_type = 'USER-DEFINED' then r.type_udt_name else r.data_type end as dtd_identifier, r.external_language as routine_body,r.routine_definition, r.sql_data_access, r.security_type, r.is_deterministic, d.description as routine_comment from information_schema.routines r left join pg_namespace ns on r.routine_schema = ns.nspname left join pg_proc p on r.specific_name = p.proname || '_' || p.oid left join pg_description d on d.objoid = p.oid where r.routine_schema = :schema'
ERROR: column p.proisagg does not exist
  Hint: Perhaps you meant to reference the column "p.prolang".
  Position: 97

mishraomp:
schemaspy is now fixed with last commit :), I will put the fixes into quickstart, please merge whenever you are ready @barrfalk or @afwilcox

afwilcox (changes requested):
@mishraomp I believe the issue with the workflow run of schemaspy is related to the default postgres password... this won't work outside of the local developer environment.  The user and password should be pulled from an Openshift Secret (See openshift.deploy.yml)

mishraomp:
> @mishraomp I believe the issue with the workflow run of schemaspy is related to the default postgres password... this won't work outside of the local developer environment. The user and password should be pulled from an Openshift Secret (See openshift.deploy.yml)

@afwilcox we don't use anything from OpenShift in the GHA workflow, it is using containers inside github, as it spins up postgis and flyway container inside the Action workflow. look here https://github.com/bcgov/nr-compliance-enforcement/blob/fix/schema-spy/.github/workflows/schemaspy.yaml#L15-L26

afwilcox (approved):
@mishraomp - I'm going to approve these changes, and we'll come back to pulling out the postgres connection information.   if this does fix the problem, then it looks like maybe those fields aren't required, as that account connection information won't work outside of a local developer workstation.

mishraomp:
> @mishraomp - I'm going to approve these changes, and we'll come back to pulling out the postgres connection information. if this does fix the problem, then it looks like maybe those fields aren't required, as that account connection information won't work outside of a local developer workstation.

it does work @afwilcox , look here https://bcgov.github.io/nr-compliance-enforcement/schemaspy/public/index.html
schemaspy works by spinning containers in the github workflow itself, so the current configuration is needed and valid. let me know if you would like to have a discussion on this one.

PR #271, opened 2024-01-30
feature/CE-39 - Add animal outcome

marqueone-ps:
# Description

Enables an officer to add an animal to wildlife conflict outcome
Story # CE-39

# How Has This Been Tested?
No cypress testing required just yet 

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-271-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-271-frontend.apps.silver.devops.gov.bc.ca/) available


afwilcox (changes requested):
1. There is something odd with all the drug text fields.   If you start typing you can't get rid of the first character.

2. Dropdowns should always source from the short description (Noticible in drugs)

3. Amount used shouldn't be a number as they want to be able to free type this (e.g. 100CC, 100ml, 100mg).  I had previously made them aware that this means that they won't be able to report on this field.

4. I can enter values into the amount discarded and discard method fields if a value other than discarded is selected.  In general I think they want these fields invisible untill it is selected to be consistent with other areas in the application.

5. Discard method and Adverse Recations has incorrect placeholder text as they are not a number field (I don't see any placeholder text in the story, so you'll need a conversation with the Designers here)

6. Missing heading "Drugs" (wireframe shows both drug and drugs depending on amount of drugs added but lets keep it simple and just have one heading "Drugs"

7. When multiple drugs exist, the drug fields should be equally spaced between the lines

8. Default Officer assigned to complaint cannot be changed in drop down in Drug Section (works on outcome)

9. The label for the officer field should be "Officer" not "Officer Assigned"

10. I don't see anywhere that states that ear tag has to be a number.   It's possible it could be something like 1A2B3C, even if it is confirmed by the business to be a number we probably want don't want to use the input type = number because users ALWAYS complain about the number spinners, because they aren't helpful.

afwilcox (approved):

PR #276, opened 2024-02-07
Fix saving complaints with coordinates - Bug/CE-447

dmitri-korin-bcps:
# Description

When attempting to create a complaint with coordinates the application displays a toast “Unable to create complaint” and fails to save the complaint data.

The issue occurs because on the backend PostGIS fails to create geometry object from the input JSON if the object type is not passed.  To fix the issue I modified a front-end component to have the object type in addition to coordinates.

Fixes # (CE-447)

# How Has This Been Tested?

- [X] Click to create a new complaint. Enter all mandatory fields. Enter coordinates (X = -123.470769, Y = 48.500560). Click Save Changes. The result is complaint created, with a pin on the map matching the entered coordinates.

## Checklist

- [x] I have added a new Cypress test _hwcr-details-create-enter-coordinates.cy.ts_ that prove my fix is effective. It passed when ran locally. 

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-276-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-276-frontend.apps.silver.devops.gov.bc.ca/) available

### DIFF
--- a/frontend/cypress/e2e/hwcr-details-create-enter-coordinates.cy.ts
+++ b/frontend/cypress/e2e/hwcr-details-create-enter-coordinates.cy.ts
@@ -1,0 +1,190 @@
+/*
+Test to verify that the user is able to click the edit button
+on the wildlife contacts details page and see all the inputs
+*/
+describe("Complaint Create Page spec - Enter Coordinates - Create View", () => {
+  const createCallDetails = {
+    description:
+      "Calling to report a black bear getting into the garbage on a regular basis. Also wanted to confirm that residents of the trailer home park could call to report sightings themselves ---- testing",
+    location: "2975 Jutland Rd.",
+    locationDescription: "---- testing",
+    incidentDateDay: "01",
+    attractants: ["Livestock", "BBQ", "Beehive"],
+    attractantCodes: ["LIVESTCK", "BBQ", "BEEHIVE"],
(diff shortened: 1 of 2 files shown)


afwilcox (changes requested):
I'm not sure we need a whole new test for this.   I think you can just add the coordinates to allegation-details-create.cy.ts and hwcr-details-create.cy.ts

    cy.get("#comp-details-edit-x-coordinate-input").click({ force: true });
    cy.get("#comp-details-edit-x-coordinate-input").clear().type(createCallDetails.xCoord);

    cy.get("#comp-details-edit-y-coordinate-input").click({ force: true });
    cy.get("#comp-details-edit-y-coordinate-input").clear().type(createCallDetails.yCoord);

dmitri-korin-bcps:
We can add coordinates to the existing tests. However, these tests also accept an address via “Location” field.  The app then passes the address to Geocoder and gets the coordinates. If we provide both an address and coordinates the app will bypass address extraction/validation and just use the coordinates we passed. I guess we may still want to preserve the test case when coordinates are obtained from the address and have a separate test with input coordinates. Let me know if otherwise.

afwilcox (approved):
Okay thanks that makes sense.

PR #301, opened 2024-03-05
Feat/ce 310

cnesmithsalus:
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-301-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-301-frontend.apps.silver.devops.gov.bc.ca/) available


afwilcox:
@cnesmithsalus - I'm able to introduce a bug via the following:
- Add 2 Animals
- Delete Animal 1
- Add a new animal 
- Delete Animal 1

At this point both animals 1 and 2 are removed

cnesmithsalus:
> @cnesmithsalus - I'm able to introduce a bug via the following:
> 
> * Add 2 Animals
> * Delete Animal 1
> * Add a new animal
> * Delete Animal 1
> 
> At this point both animals 1 and 2 are removed

Back to you once github actions done

frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/animal-outcome-item.tsx
```diff
@@ -16,7 +16,7 @@ import { selectOfficersByAgencyDropdown } from "../../../../../store/reducers/of
 import { formatDate, getAvatarInitials } from "../../../../../common/methods";
 import { from } from "linq-to-typescript";
 import { DrugItem } from "./drug-item";
-import { BsPencil } from "react-icons/bs";
+import { BsPencil, BsTrash } from "react-icons/bs";
```
Scarlett-Truong, line 19: I think the BsTrash3 looks more exact in the design

frontend/src/app/components/containers/complaints/outcomes/hwcr-outcome-by-animal.tsx
```diff
@@ -40,6 +40,19 @@ export const HWCROutcomeByAnimal: FC = () => {
...
+    updatedList.forEach(animal => {
+      if(animal.id > id)
+      {
+        animal.id = id;
+      }
+    });
+    setAnimals(updatedList);
+    setShowForm(false);
+  };
+
```
Scarlett-Truong, line 55: I don't think using `filter` is a good idea cuz you want to completely remove it. There are some weird behaviours you may encounter, for ex this bug: add 3 animals -> delete the first one -> delete the second one -> the third one supposed to be there but it's gone as well.

I'm using `splice` instead, it's more straightforward and no need for a second loop through the array.

Also I'm not sure if we need the warning modal popup before deleting, cuz in my ticket it's said there in the requirement, but in yours is not. Should check again with @afwilcox

frontend/src/assets/sass/hwcr-animal-outcome.scss
```diff
@@ -111,6 +111,41 @@
     border-radius: 3px;
     padding: 24px;
 
+    .animal-item-edit {
```
Scarlett-Truong, line 114: I remembered removing this class when doing the edit button cleanup 😆 I don't think it's using elsewhere

frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/animal-outcome-item.tsx
cnesmithsalus: Agree

frontend/src/app/components/containers/complaints/outcomes/hwcr-outcome-by-animal.tsx
cnesmithsalus: This one I still have to iterate after the filter/splice since I have to reset the key if it's after the key I'm removing is gone - so I don't think there's an advantage over one or the other. My condition to reset the key was animal.id = id; instead of animal.id = animal.id - 1; which was dumb though

frontend/src/assets/sass/hwcr-animal-outcome.scss
cnesmithsalus: I think it'll affected some of the other button icons when I tried to just get rid of it

frontend/src/app/components/containers/complaints/outcomes/hwcr-outcome-by-animal.tsx
afwilcox: @cnesmithsalus  - think you should use @Scarlett-Truong's way instead.  It seems to work without needing the iteration.   Also I found another issue with your PR that doesn't exist on hers:

If you press delete while you are adding or editing another one it will close it.   I'm probably going to merge her PR into main first so maybe hold tight...  I've got some more questions to Neil about some of the requirements.

afwilcox:
@cnesmithsalus - One regression item... after you add two ear tags the 'Add Ear Tag' button should no longer be available

One new item... the label on the button for the confirmation says "Equipment" instead of "animal outcome"

cnesmithsalus:
@afwilcox Regression fixes complete

afwilcox (approved):

PR #302, opened 2024-03-06
feat: CE-510 frontend assessment information

barrfalk:
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist


Scarlett-Truong:
@barrfalk correct me if I was wrong, I thought after clicking the Save button, it should appeared like this as in the design 
Screen Shot 2024-03-06 at 12 13 15 PM

afwilcox:
In addition to what @Scarlett-Truong mentioned above I'll point out two other things:

1. Because you are modifying the state you get the prize of being the first person to do so using the state migrations.   Talk to @marqueone-ps  if you have questions but there is a page on the wiki and I think the code is well commented.
2. When you are in the validation state the top row is out of alignment:

image

barrfalk:
> @barrfalk correct me if I was wrong, I thought after clicking the Save button, it should appeared like this as in the design Screen Shot 2024-03-06 at 12 13 15 PM

Thanks @Scarlett-Truong !  I wasn't quite finished with this ticket, but I'll be adding that in!

dmitri-korin-bcps:
@barrfalk 
Is this still WIP?
In /src/app/store/reducers/outcomes/assessment.ts there is a function to get assessment from the backend. I think you may also need to do some preparation to call backend to save/update assessment.

afwilcox:
@dmitri-korin-bcps - do you want to merge your backend changes into this PR and this can have both the frontend and backend changes?  I'm good with the changes @barrfalk made on the front end (I'm going to raise a bug for alignment on error issue as it exists in more than one component.

dmitri-korin-bcps:
@afwilcox 
Good idea. I plan to work on Complaint Management backend tomorrow when I finish refactoring for Case Management part.

frontend/src/app/store/reducers/outcomes/assessment.ts
```diff
@@ -0,0 +1,67 @@
...
+export const getAssessment =
+  (complaint_identifier?: string): AppThunk =>
+  async (dispatch) => {
+    const parameters = generateApiParameters(
+      `${config.API_BASE_URL}/v1/case/${complaint_identifier}`
+    );
+    const response = await get<Assessment>(dispatch, parameters);
+
+    if (response) {
+      dispatch(
```
marqueone-ps, line 58: you're going to want to do this instead here: 
```
      dispatch(
        setAssessment(response)
      );
```

frontend/src/app/store/reducers/outcomes/assessment.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+};
+
+export const assessmentSlice = createSlice({
+  name: "assessment",
+  initialState,
+  // The `reducers` field lets us define reducers and generate associated actions
+  reducers: {
+    setAssessment: (state, action) => {
+      const { payload } = action;
```
marqueone-ps, line 24: Here you just want to spread your assessment instead of creating an assessment object so you'd do somethign like

const { payload: { assessment } } = action
return { ...assessment }

frontend/src/app/store/reducers/outcomes/assessment.ts
```diff
@@ -0,0 +1,67 @@
...
+      return initialState;
+    });
+  },  
+});
+
+// export the actions/reducers
+export const { setAssessment } = assessmentSlice.actions;
+
+export const selectAssessment = (state: RootState): Assessment => {
+  return state.assessment.assessment;
```
marqueone-ps, line 43: Same thing here, when simplifiying your state, you can do the following here: 

```
const { assessment } = state;
return assessment
```

marqueone-ps:
@barrfalk you're going to want to make sure you add a redux migration to handle adding that new assessment state property. Additionally you'll want to blacklist that property in the redux-persist config as well so that we don't end up persisting that data.

In your initial state, you don't need to include the properties that are undefined in the assessment type. In your Assessment type I can see a couple things you may want to remove and clean up as well. In fact it looks like you're returning the whole entity rather than the model from the server, and the audit properties are redundant unless they're actually being used. You've also got a case miss match snake_case vs camelCase and we really should be using camelCase. 

The other thing I see is the types on the Assessment type. You don't want to use the Option type, instead for officer you'd want to use a UUID/string and for justification you'd want to use string. 

Your reducer has a couple problems as well, not huge but still a bit of a problem. your assessment state is doubled up so when we look at it in redux it would look like this 
```
{
   complaints,
   code-tables,
   assessment: { 
      assessment: { 
      ....
  }
}
``` 
what you really want is this: 
```
{
   complaints,
   code-tables,
   assessment: { 
      ....
  }
}
```

afwilcox:
Looks good @dmitri-korin-bcps I only found one minor issue if you could take a look at this, and also the code smells I think this one is good to go.   Most of them look easy, the complexity one is the toughest - you might need to break it down into edit and insert functions.   

Action Required = 'Yes' does not seem to be working properly.  It looks good when you save it, but when you come back it displays 'No'

I'll fix the unit tests.

marqueone-ps:
@dmitri-korin-bcps that complexity item that sonar cloud flagged is definitely going to be a bear to deal with 

```
if (!createAssessmentInput.createAssessmentInput.assessmentDetails.actions.map((action) => {
              return action.actionCode
            }).includes(item.assessmentType))
``` 
for example this you may want to create an inline function instead of trying to map and includes in that if statement. I'd also recommend you do some object destructuring as you're really far down the dot notation rabbit hole here.

dmitri-korin-bcps:
@afwilcox 

I fixed the issue with Action Required. The code change is in Case Management.

afwilcox:
Looks better now thanks @dmitri-korin-bcps  - I found one more issue:  it looks like you are passing the display name in the create/update user fields instead of the user name.  Hopefully this is a quick fix.

dmitri-korin-bcps:
@afwilcox and @marqueone-ps 

The code smells, including the complexity one is now addressed. I extracted separate functions for the addition and update of assessments, did some refactoring to remove ifs and used object destructuring to make the code more readable.

Also earlier I committed the code to use idir_username as a create/update user.

afwilcox (approved):

PR #312, opened 2024-03-14
bug: CE-491 Adding Jetstream prevent duplicate complaints

barrfalk:
# Description

The following only occurs in the test environment.   It is likely a result of multiple webEOC containers running.   

Steps to Reproduce:

    Consume complaints from webEOC

Observed Behavior:

    Some complaints have duplicate entries in the staging table 

    Both complaints have a status of ‘Error’ however the complaint was loaded into the system

Expected Behavior:

    If a complaint was loaded into the system there should be a row with a status of ‘SUCCESS’

    Multiple error rows for a complaint are okay, however it looks like the rows are being created at the same time, which is probably because the message is going to both webeoc containers.

Fixes CE-491

# How Has This Been Tested?

- [x] Created 2 WebEOC replicas, verified that there are no errors in the staging table

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-312-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-312-frontend.apps.silver.devops.gov.bc.ca/) available

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-312-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-312-frontend.apps.silver.devops.gov.bc.ca/) available


afwilcox:
I am still seeing this error on this PR.   By looking at the staging table:

image

webeoc/src/webeoc-complaints-scheduler/webeoc-complaints-scheduler.service.ts
```diff
@@ -124,13 +113,13 @@ export class WebEOCComplaintsScheduler {
...
-      this.logger.error('Error fetching complaints from WebEOC:', error);
+      this.logger.error("Error fetching complaints from WebEOC:", error);
       throw error;
     }
   }
 
   // Formats a date to be in a format required by WEBEOC.  Used to determine how far back in time to go to grab new complaints.
   private formatDate(date: Date): string {
-    return date.toISOString().replace('T', ' ').substring(0, 19);
+    return date.toISOString().replace("T", " ").substring(0, 19);
```
marqueone-ps, line 123: This doesn't seem like a good idea here. A better solution would be to use datefns to format your date. It shouldn't happen but this feels like it has the potential to explode and because its happening outside of user control the chances of catching it could be difficult.

webeoc/src/staging-complaints-api-service/staging-complaints-api-service.service.ts
```diff
@@ -19,39 +18,35 @@ export class StagingComplaintsApiService {
...
     try {
-      this.logger.debug(
-        'Creating new complaint based on new complaint from webeoc.',
-      );
+      this.logger.debug("Creating new complaint based on new complaint from webeoc.");
       const apiUrl = `${process.env.COMPLAINTS_MANAGEMENT_API_URL}/staging-complaint/process/${complaint_identifier}`;
       this.logger.debug(`Posting new complaint. API URL: ${apiUrl}`);
 
       // Include the API key in the request headers
       const config = {
```
marqueone-ps, line 41: I'd recommend turning this: 
```
      const config = {
        headers: {
          'x-api-key': process.env.COMPLAINTS_API_KEY, // Ensure this environment variable is set
        },
      };
```
into this:
```
  private readonly _apiConfig = { 
    headers: { 
      "x-api-key": process.env.COMPLAINTS_API_KEY
    }
  }
``` 
into a reusable private object as you're using this object in a few other locations. in the event that you need to add a header, or additional config value you can spread your source config object into a new local config.

marqueone-ps (changes requested):

webeoc/src/webeoc-complaints-scheduler/webeoc-complaints-scheduler.service.ts
barrfalk: Good find, I'll make that change.

webeoc/src/staging-complaints-api-service/staging-complaints-api-service.service.ts
barrfalk: Good find, I'll make that change.

barrfalk: Fixed

webeoc/src/webeoc-complaints-scheduler/webeoc-complaints-scheduler.service.ts
barrfalk: Fixed

marqueone-ps (approved):

afwilcox:
@barrfalk - Sorry no dice here, this doesn't look like it's resolved.   See Complaint 24-000195 in Openshift Database

afwilcox (approved):

PR #315, opened 2024-03-18
feat/CE-187-New dropdown available without clearing cache

Scarlett-Truong:
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-315-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-315-frontend.apps.silver.devops.gov.bc.ca/) available


afwilcox:
@Scarlett-Truong  - This looks good so far, the only thing I see is that I think you need to use a Redux migration for the addition of the code table version into the state to ensure that it gets loaded properly the first time.

I'm going to do a bit more testing, but so far it's working exactly as I would have expected it to.

@marqueone-ps  can answer any questions about it if you need it.

backend/src/v1/configuration/configuration.controller.ts
```diff
@@ -22,8 +24,32 @@ export class ConfigurationController {
...
+    @Token() token) 
+  {
+    const result = await this.configurationService.findOne(configurationCode);
+
+    //If configuration is code table version, call another case managment api
+    if(configurationCode === 'CDTABLEVER') {
+      const { data } = await get(token, { 
+        query : '{getConfigurationCode(configuration_code: "CDTABLEVER"){configuration_code configuration_value  active_ind}}'
+      });
+      const caseData = {
```
marqueone-ps, line 38: @Scarlett-Truong you might want to check that result is both not null and that there are elements in the results array. There may be a scenario where that can cause an error when creating complaintData.

frontend/src/app/store/reducers/app.ts
```diff
@@ -389,6 +414,10 @@ const initialState: AppState = {
   configurations: {
     configurations: undefined,
   },
+  codeTableVersion: (localStorage.getItem("codeTableVersion")!== null) ?
```
marqueone-ps, line 417: @Scarlett-Truong is this necessary? Is there a reason the codeTableVersion would live outside the scope of redux and have a separate localState property for tracking the codeTableVersion?

backend/src/v1/configuration/configuration.controller.ts
Scarlett-Truong: yeah, good point 😃

frontend/src/app/store/reducers/app.ts
marqueone-ps: Thats why I'm a bit confused by this. You're checking the localState for that value and setting it to the initialState, but if that value isn't there you're defaulting it to ```{ complaintManagement: "1", caseManagement: "1" } ``` and then once you're request for the codeTables is complete you're dispatching the version numbers to redux. The localStorage step seems redundant.

PR #319, opened 2024-03-24
Ce 365 v2

cnesmithsalus:
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist


afwilcox:
@cnesmithsalus  - Looks good to me, however I did notice that the dropdown options were changed after I added the technical design so you'll need to refresh these.   Unfortunately they put one in the middle of the other ones so we will need to reindex the display order :(

afwilcox:
@cnesmithsalus  - Also I don't think you'll be able to handle the duplication issue in time, but the two code smells seem like they can be fixed.   For the duplicate function, I think it should be renamed to findCase instead of findAssessment since the whole case object is returned.   Then it can be called in both places.

afwilcox (approved):

PR #323, opened 2024-04-01
feature: CE-554

cnesmithsalus:
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

### DIFF
--- a/frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/add-drug.tsx
+++ b/frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/add-drug.tsx
@@ -45,6 +45,16 @@ export const AddDrug: FC<props> = ({
 
   const [showDiscarded, setShowDiscarded] = useState(remainingUse === "DISC");
 
+  const [vialNumberErrorMessage, setVialNumberErrorMessage] = useState(!vial ? "Required" : "");
+  const [drugNameErrorMessage, setDrugNameErrorMessage] = useState(!drug ? "Required" : "");
+  const [amountUsedErrorMessage, setAmountUsedErrorMessage] = useState(amountUsed <= -1 ? "Required" : "");
+  const [injectionMethodErrorMessage, setInjectionMethodErrorMessage] = useState(!injectionMethod ? "Required" : "");
+
+  function isPositiveInt(str: string) {
+    var n = Math.floor(Number(str));
+    return (str !== "") ? n !== Infinity && String(n) === str && n >= 0 : true;
+}
+
   const updateModel = (property: string, value: string | Date | number | null | undefined) => {
     const source = {
       id,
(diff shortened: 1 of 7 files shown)


afwilcox:
@cnesmithsalus I saw a couple of UX items that I think we should adjust (unless you were told to explicitly do any of these:
- If only the species is entered we should drop the comma at the end as it looks silly
- If there is no outcome entered we should hide the officer initial placeholder
- the required fields show as soon as you select an outcome (before pressing save)

image

image

frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/add-drug.tsx
```diff
@@ -79,10 +88,12 @@ export const AddDrug: FC<props> = ({
   };
 
   const handleAmountUsed = (input: string) => {
+    if(isPositiveInt(input))
```
marqueone-ps, line 91: I don't think this is a good idea for a couple reasons. The big one is that units can be decimals, 1.5, 2.5 etc, and this effectively prevents someone from using decimals. Additionally in the edge case where someone uses a negative value, it would be easier to flip the negative to a positive value

frontend/src/app/components/containers/complaints/outcomes/hwcr-prevention-education.tsx
```diff
@@ -288,6 +288,7 @@ export const HWCRComplaintPrevention: FC = () => {
                         className="comp-details-edit-calendar-input" // Adjust class as needed
                         classNamePrefix="comp-select" // Adjust class as needed
                         errMsg={preventionDateErrorMessage} // Pass error message if any
+                        maxDate={new Date()}
```
marqueone-ps, line 291: If you want to pass a date down to the child components like this, you will want to create a top level ```const currentDate: Date = new Date()``` and pass it down instead of a new date, this way the date is consistent for everything in the component.

marqueone-ps (changes requested):
There's a small problem in the `hwcr-outcome-by-animal.tsx` component where the component is selecting black bear for the species when it should only default to black bear if there is no species

frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/add-drug.tsx
cnesmithsalus: Agree about the decimal, for some reason I thought this was only ints. Disagree about flipping the value, think that could be confusing for the user and just preventing the negative input is more intuitive.

frontend/src/app/components/containers/complaints/outcomes/hwcr-prevention-education.tsx
cnesmithsalus: Makes sense

frontend/src/app/components/containers/complaints/outcomes/animal-outcomes/add-drug.tsx
marqueone-ps: if you don't want to flip the negative to a positive then you'll need to provide a message to the user that they can't use a negative value otherwise it will be very confusing to the user why they can't add a value there.

marqueone-ps (approved):

afwilcox:
@cnesmithsalus - Sorry found another issue :(

- Expand Drugs
- Fill in Everything except the officer and date
- Press Save

This should not let me save, but it does.

cnesmithsalus:
@afwilcox I want to throw my computer out of the window -- I *think* it should be good to go now

afwilcox (approved):
@cnesmithsalus - I think we finally got it!

PR #329, opened 2024-04-05
fix: update redux migration for File Review

Scarlett-Truong:


frontend/src/app/store/migrations/migration-5.ts
```diff
@@ -0,0 +1,17 @@
+//Add personGuid in app and isReviewRequired,reviewComplete in cases
+export const Migration5 = {
```
marqueone-ps, line 2: I would recommend not using this as a name for your migration. The fist example I crated did this as an example. You really should use a name that helps explain the migration at a glance

marqueone-ps (changes requested):
Just the one small change for the migration name

marqueone-ps (approved):

afwilcox (approved):

PR #411, opened 2024-05-01
feat: CE-636 allow user to scroll through map

Scarlett-Truong:
---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-411-backend.apps.silver.devops.gov.bc.ca/) available
[Frontend](https://nr-compliance-enforcement-411-frontend.apps.silver.devops.gov.bc.ca/) available

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -61,6 +61,7 @@
         "jwt-decode": "^3.1.2",
         "keycloak-js": "^21.0.0",
         "leaflet": "^1.9.4",
+        "leaflet-gesture-handling": "^1.2.2",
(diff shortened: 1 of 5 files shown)


marqueone-ps (approved):
works like a charm!

Not a change request more of a question, should the css import be in the gesture component, or would it be better placed in with the sass files?

Scarlett-Truong:
@marqueone-ps ikr 
For your css question, I prefer to keep thing co-location, so any css related to a component will be closed to each other, in case if we want to remove the component, we can just delete the file 🤓

afwilcox (approved):

PR #420, opened 2024-05-13
feature/CE-629-Implement-changes-to-Complaint-Assessment-section-based-on-usability-testing-feedback

dmitri-korin-bcps:
# Description

This change is required to address usability issue. The officers want it to be simple to close a HWC when no action is required so that they are not wasting time filling out fields that don’t make sense for certain use cases.

Fixes # (CE-629)

# How Has This Been Tested?

- [x]  In the complaint assessment the option to select ‘Action required’ appears above other options. Action required, officer and date fields are mandatory. If Action required is selected as no, Justification field is required. If Action required field is selected as yes, at least one checkbox must be selected (is mandatory)

## Checklist

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-420-frontend.apps.silver.devops.gov.bc.ca/api/) available
[Frontend](https://nr-compliance-enforcement-420-frontend.apps.silver.devops.gov.bc.ca/) available

### DIFF
--- a/frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
+++ b/frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
@@ -33,6 +33,7 @@ import { CompTextIconButton } from "../../../common/comp-text-icon-button";
 import "../../../../../assets/sass/hwcr-assessment.scss";
 import { selectAssessment } from "../../../../store/reducers/case-selectors";
 import { getAssessment, upsertAssessment } from "../../../../store/reducers/case-thunks";
+import { OptionLabels } from "../../../../constants/option-labels";
 
 export const HWCRComplaintAssessment: FC = () => {
   const dispatch = useAppDispatch();
@@ -230,7 +231,7 @@ export const HWCRComplaintAssessment: FC = () => {
           value: selectedJustification?.value,
         },
         assessment_type:
-          selectedActionRequired?.label === "No"
+          selectedActionRequired?.label === OptionLabels.OPTION_NO
             ? []
             : selectedAssessmentTypes?.map((item) => {
                 return {
@@ -280,7 +281,10 @@ export const HWCRComplaintAssessment: FC = () => {
       hasErrors = true;
     }
 
-    if (selectedActionRequired?.value === "Yes" && (!selectedAssessmentTypes || selectedAssessmentTypes?.length <= 0)) {
+    if (
+      selectedActionRequired?.value === OptionLabels.OPTION_YES &&
+      (!selectedAssessmentTypes || selectedAssessmentTypes?.length <= 0)
+    ) {
       setAssessmentRequiredErrorMessage("One or more assessment is required");
       hasErrors = true;
     }
--- a/frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
+++ b/frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
@@ -229,12 +229,15 @@ export const HWCRComplaintAssessment: FC = () => {
           key: selectedJustification?.label,
           value: selectedJustification?.value,
         },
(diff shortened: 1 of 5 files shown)


afwilcox:
@dmitri-korin-bcps - I believe the cypress error reported here is legitimate as the tests will require some updates based on the new workflow

barrfalk (approved):

afwilcox (changes requested):
@dmitri-korin-bcps - I think the third cypress test (cancelling assessment edits) still needs an update.   It will work on a clean database, but not on a subsequent run, as the checkboxes aren't visible (as the last test sets the Action required to "No")

It might be easier to update the last test to not set the Action Required to "No" as this will ensure the checkboxes are still visible

frontend/src/app/store/reducers/case-thunks.ts
```diff
@@ -98,14 +98,16 @@ const addAssessment =
...
           actionNotRequired: assessment.action_required === "No",
-          actions: assessment.assessment_type.map((item) => {
-            return {
-              date: assessment.date,
-              actor: assessment.officer?.value,
-              activeIndicator: true,
-              actionCode: item.value,
-            };
-          }),
+          actions: assessment.assessment_type
```
marqueone-ps, line 101: I don't think the ternary statement here isn't necessary. If `assessment_type` is an empty array `map()` will return an empty array

frontend/src/app/store/reducers/case-thunks.ts
```diff
@@ -166,14 +168,16 @@ const updateAssessment =
...
           actionJustificationCode: assessment.justification?.value,
-          actions: assessment.assessment_type.map((item) => {
-            return {
-              actor: assessment.officer?.value,
-              date: assessment.date,
-              actionCode: item.value,
-              activeIndicator: true,
-            };
-          }),
+          actions: assessment.assessment_type
```
marqueone-ps, line 171: The same thing as with line 102, the ternary statement isn't needed here

frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
```diff
@@ -229,12 +229,15 @@ export const HWCRComplaintAssessment: FC = () => {
...
           value: selectedJustification?.value,
         },
-        assessment_type: selectedAssessmentTypes?.map((item) => {
-          return {
-            key: item.label,
-            value: item.value,
-          };
-        }),
+        assessment_type:
+          selectedActionRequired?.label === "No"
```
marqueone-ps, line 233: I think that if we're using "Yes" and "No" values these should be treated as constants so that they can be reused else where, makes things less error prone

frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
```diff
@@ -277,7 +280,7 @@ export const HWCRComplaintAssessment: FC = () => {
       hasErrors = true;
     }
 
-    if (!selectedAssessmentTypes || selectedAssessmentTypes?.length <= 0) {
+    if (selectedActionRequired?.value === "Yes" && (!selectedAssessmentTypes || selectedAssessmentTypes?.length <= 0)) {
```
marqueone-ps, line 283: same comment as line 233

frontend/src/app/store/reducers/case-thunks.ts
```diff
@@ -98,14 +98,16 @@ const addAssessment =
         caseCode: "HWCR",
```
marqueone-ps, line 98: There should be a constant to use here instead of hardcoding the HWCR value

marqueone-ps (changes requested):
Just a couple small changes, otherwise everything looks good from a code perspective :)

afwilcox (approved):
Remaining Failing Unit tests are a result of the test account being unavailable.

PR #423, opened 2024-05-16
feat: CE-99 Receive Updates to HWCs and ERSs from WebEOC

barrfalk:
# Description

Receive Updates to HWCs and ERSs from WebEOC as per CE-99

Fixes # (issue)

# How Has This Been Tested?

 - [x] Updates from WebEOC to the following fields are displayed with the date and time of update, as seen in designs: Complaint details, Complaint location, Location description, X coordinate, Y coordinate
 - [x] Only the updated information is displayed (do not show the fields that have not changed) 
 - [x] Updates to the remaining fields (not included in the above list) overwrite the information in the NatCom Complaint details 
 - [x] If the location information changes, the pin on the map will remain at the original location (until more functionality is implemented in the future to handle multiple pins) 
 - [x] The list view will continue to display the original complaint information (complaint details, complaint location, location description) 
 - [x] The Last updated field of a complaint is updated to account for information coming in from WebEOC as an update.
 - [x] This is correctly reflected in the list view and the complaint details. 
 - [x] If an update is received with no information, do not display the update 
 - [x] An Update should only be recorded if it meets both of the following criteria
 - [x] It is different from the current value for a field - to prevent unnecessary updates
 - [x] It is different from the original value for a field  - to prevent accidental reversions of updates due to webEOC default behaviour

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-423-frontend.apps.silver.devops.gov.bc.ca/api/) available
[Frontend](https://nr-compliance-enforcement-423-frontend.apps.silver.devops.gov.bc.ca/) available


backend/src/types/models/complaint-updates/complaint-updates.ts
```diff
@@ -0,0 +1,15 @@
+import { Point } from "typeorm";
+
+export class ComplaintUpdatesDto {
```
marqueone-ps, line 3: for your properties unless you know there's going to be a null value you may want to use the `?` operator on the type instead.  This will help with when we run into silly null problems, plus it will kick vscode to remind you the property you're using could be undefined.

backend/src/v1/complaint/complaint.controller.spec.bak
marqueone-ps, line 1: Is this file required any more?

backend/src/v1/staging_complaint/staging_complaint.controller.ts
```diff
@@ -14,15 +15,31 @@ import { Public } from "../../auth/decorators/public.decorator";
 export class StagingComplaintController {
   constructor(private readonly stagingComplaintService: StagingComplaintService) {}
 
-  @Post()
+  @Post("/creates")
```
marqueone-ps, line 18: is this a typo? ```@Post("/creates")``` vs ```@Post("/create")```
similar question about ```@Post("/updates")``` vs ```@Post("/update")```

backend/src/v1/staging_complaint/staging_complaint.controller.ts
```diff
@@ -14,15 +15,31 @@ import { Public } from "../../auth/decorators/public.decorator";
...
 
   @Post("/process/:complaintIdentifier")
   @UseGuards(ApiKeyGuard)
-  async process(@Param("complaintIdentifier") complaintIdentifier: string): Promise<any> {
-    return await this.stagingComplaintService.process(complaintIdentifier);
+  async processWebEOCComplaint(@Param("complaintIdentifier") complaintIdentifier: string): Promise<any> {
+    return await this.stagingComplaintService.processWebEOCComplaint(complaintIdentifier);
+  }
+
+  @Post("/process/:complaintIdentifier/:update_number")
```
marqueone-ps, line 36: I know this seems like nitpicking: ```@Post("/process/:complaintIdentifier/:update_number")``` which case_type are you using here, camelCase, or snake_case. should choose one instead of both

backend/src/v1/staging_complaint/staging_complaint.service.ts
```diff
@@ -53,9 +59,56 @@ export class StagingComplaintService {
...
+    }
+
+    const newStagingComplaint = this.stagingComplaintRepository.create();
+    newStagingComplaint.stagingStatusCode = { stagingStatusCode: StagingStatusCodeEnum.PENDING } as StagingStatusCode;
+    newStagingComplaint.stagingActivityCode = {
+      stagingActivityCode: StagingActivityCodeEnum.UPDATE,
+    } as StagingActivityCode;
+    newStagingComplaint.complaintIdentifer = stagingComplaint.parent_incident_number;
+    newStagingComplaint.complaintJsonb = stagingComplaint;
+    newStagingComplaint.createUserId = "WebEOC";
```
marqueone-ps, line 93: minor suggestion here
```
const userId = "WebEOC";
const current = new Date();

newStagingComplaint.createUserId = userId
newStagingComplaint.createUtcTimestamp = current
newStagingComplaint.updateUserId = userId
newStagingComplaint.updateUtcTimestamp = current
```

rational for this is that if we need to change the user, there's only one place to do it vs two, and it keeps the create and update dates inline

frontend/src/app/components/containers/complaints/webeoc-complaint-updates/webeoc-complaint-update-list.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const toggleExpand = (id: string) => {
+    setExpandedUpdates((prev) => ({
+      ...prev,
+      [id]: true,
+    }));
+  };
+
+  useEffect(() => {
+    Object.keys(descriptionRefs.current).forEach((id) => {
```
marqueone-ps, line 30: nice! :)

frontend/src/app/components/containers/complaints/webeoc-complaint-updates/webeoc-complaint-update-list.tsx
```diff
@@ -0,0 +1,128 @@
...
+          ...prev,
+          [id]: true,
+        }));
+      }
+    });
+  }, [complaintUpdates]);
+
+  return (
+    <>
+      {complaintUpdates && complaintUpdates.length > 0 ? (
```
marqueone-ps, line 43: I would avoid a ternary statement here and instead suggest using a short circuit instead so as to avoid adding the "" to the markup

 ```
 {complaintUpdates && complaintUpdates.length !== 0 && (<>component goodness</>)}
 ```
If the ternary is needed, in your else block instead of an empty string use a fragment. This will prevent the empty string from being rendered
```
 {complaintUpdates && complaintUpdates.length !== 0 ? (<>component goodness</>) : <></> }
```

webeoc/src/webeoc-complaints-scheduler/webeoc-complaints-scheduler.service.ts
```diff
@@ -98,28 +82,44 @@ export class WebEOCComplaintsScheduler {
...
+        items: [dateFilter],
       },
     };
 
-    const url = `${process.env.WEBEOC_URL}/board/Conservation Officer Service/display/List - API COS Complaints Incidents`;
-
     try {
       const response = await axios.post(url, body, config);
-      return response.data as Complaint[];
+      return response.data as any[];
```
marqueone-ps, line 91: I would avoid using `any` if you can

marqueone-ps (changes requested):
Just a few things to look at, but otherwise looks great!

frontend/src/app/components/containers/complaints/webeoc-complaint-updates/webeoc-complaint-update-list.tsx
barrfalk: Good call, I'll make it so

backend/src/v1/staging_complaint/staging_complaint.controller.ts
barrfalk: I like this question.  I agree that we need something better, but I don't know what.  The thing here is that these routes are used to Post a create record and to Post an update record.  Both are posts (not a patch for update, because it's actually creating new data).  More specifically, we're Posting either a WebEOC create or Posting a WebEOC update.  I don't like using /create or /update because those are verbs.  Maybe webeoc-complaint and webeoc-complaint-update?

marqueone-ps: Might have our lines crossed here. My question wasn't about the the verb you're using, in fact I think POST is 100% correct in either case honestly, but rather the pluralization of your action name. There's something that feels weird about "creates" over just "create"

backend/src/v1/staging_complaint/staging_complaint.controller.ts
barrfalk: Good catch.  I've standardized on camel case for parameters.

backend/src/v1/staging_complaint/staging_complaint.controller.ts
barrfalk: I agree, I don't like creates.  I also don't like create though, since it's a verb.  I'm thinking webeoc-complaint, and webeoc-complaint-update works better since we're posting either a webeoc-complaint or a webeoc-complaint-update.  I've made that change, let me know what ya think!

marqueone-ps (approved):

afwilcox (changes requested):
The webeoc scheduler will need to also consider the pilot flag for updates.   Unfortunately it's a different flag than the Incidents (current merge conflict - which should be straight forward to resolve)   The additional logic required for updates is:

{"fieldname":"flag_UPD","operator":"Equals","fieldvalue":"Yes"}

afwilcox (changes requested):
Found a couple other issues as well:

- We should be touching the complaint record when updates come in so they appear at the top of the list
- line breaks are getting dropped in the updates so if I enter multiple paragraphs it all gets condensed into one.   Do we want this?
- There is something weird going on with the coordinates.   It looks like the original complaints are coming through more precision than the updates, so the updates are flagging coordinates are being changed when they are not.   To resolve this it looks like there are two pairs of coordinates in the incidents, and we should be using the one with less precision.

afwilcox (approved):

PR #443, opened 2024-05-31
feat: CE-539 - Export complaint to PDF

marqueone-ps:
# Description

Exports a complaint (HWCR and ERS) to a PDF document via the CDOGS API

Story # CE-539

# How Has This Been Tested?

- [x] Can generate HWCR
- [x] Can Generate ERS

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-443-frontend.apps.silver.devops.gov.bc.ca/api/) available
[Frontend](https://nr-compliance-enforcement-443-frontend.apps.silver.devops.gov.bc.ca/) available


backend/src/external_api/cdogs/cdogs.service.ts
```diff
@@ -0,0 +1,192 @@
...
+  readonly baseUri: string;
+  readonly clientId: string;
+  readonly clientSecret: string;
+  readonly grantType: string;
+
+  @Inject(ConfigurationService)
+  readonly configService: ConfigurationService;
+
+  constructor() {
+    this.authApi = process.env.COMS_JWT_AUTH_URI;
```
barrfalk, line 25: I know there was a chat on slack about this, but I feel like this isn't right.  If I understand this correctly, this has a dependency on the COMS Keycloak realm.  If COMS is taken offline, this would end up breaking our ability to work with CDOGS.  

Does this work with the standard realm?

For example, rename this to JWT_AUTH_URI, and use the values https://dev.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/token, https://test.loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/token, or https://loginproxy.gov.bc.ca/auth/realms/standard/protocol/openid-connect/token

marqueone-ps: As per our slack conversation, I can confirm that the url to authorize CDOGS use is correct so we can't re-use the existing JWT_AUTH_URI but at the same time we can't use that url and just change the realm for additional common services

backend/src/external_api/cdogs/cdogs.service.ts
```diff
@@ -0,0 +1,192 @@
...
+          Authorization: `Bearer ${apiToken}`,
+        },
+      };
+
+      const documentData = await this._applyData(data, documentName);
+
+      const response = await axios.post(url, documentData, config);
+      return response;
+    } catch (error) {
+      console.log(`exception: unable to export document for complaint: ${data.id} - error: ${error}`);
```
barrfalk, line 188: We should be using the Logger middleware.

e.g. private readonly logger = new Logger(CdogsService.name);

then this.logger.error(bla)

There are a few reasons, but in this case we have the same exception text in several classes, and the console log statement doesn't tell us which class throws this error in the logs.

marqueone-ps: Thanks for catching that, I totally forgot about the logger, and you're absolutely right

backend/src/external_api/cdogs/cdogs.service.ts
```diff
@@ -0,0 +1,198 @@
...
+  //--
+  //-- render complaint to pdf
+  //--
+  generate = async (documentName: string, data: any, type: COMPLAINT_TYPE): Promise<AxiosResponse> => {
+    const templateCode = type === "HWCR" ? CONFIGURATION_CODES.HWCTMPLATE : CONFIGURATION_CODES.ERSTMPLATE;
+
+    try {
+      const apiToken = await this.authenticate();
+
+      if (!(await this.isTemplateCached(apiToken, templateCode))) {
```
barrfalk, line 174: We shouldn't log the api token because it will be viewable to anyone scraping the logs in Kibana

marqueone-ps: I actually just removed that unnecessary logging there because of just that :)

backend/src/v1/document/document.controller.ts
```diff
@@ -0,0 +1,50 @@
...
+
+      res.set({
+        "Content-Type": "application/pdf",
+        "Content-Disposition": `attachment; filename=${fileName}`,
+        "Content-Length": buffer.length,
+      });
+
+      res.end(buffer);
+    } catch (error) {
+      this.logger.log(`exception: unable to export document for complaint: ${id} - error: ${error}`);
```
barrfalk, line 46: nitpick, but this should be logger.error instead of logger.log.

barrfalk: I'll make the change though!

barrfalk (approved):

PR #454, opened 2024-06-11
feat: CE-765 WebEOC In Production

barrfalk:
# Description

WebEOC is not configured to start in production
WebEOC now has a dependency on the backend, meaning it will start up only after the backend starts.

Fixes # (issue)

# How Has This Been Tested?

This is a pipeline change, and can't be tested until migrated to test and prod.

## Checklist

## Further comments

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-454-frontend.apps.silver.devops.gov.bc.ca/api/) available
[Frontend](https://nr-compliance-enforcement-454-frontend.apps.silver.devops.gov.bc.ca/) available


.github/workflows/merge-main.yml
```diff
@@ -125,29 +123,95 @@ jobs:
...
+
+  deploys-prod:
+    name: Deploy Other Components to PROD
+    needs:
+      - deploys-prod-backend
+    environment: prod
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
-        name: [backend, database, frontend, init]
```
afwilcox, line 136: Why was backend removed from here?   I think we need it to pass in the flyway parameters?  

Also will need to resolve the merge-conflict.

afwilcox (changes requested):

.github/workflows/merge-main.yml
barrfalk: Good observation.  In the past, by using the matrix strategy, webeoc would start before the backend.  To fix this, I'm first deploying the backend first (it still has the flyway parameters, see the deploys-prod-backend job in merge-main.yml), and then deploying everything else after.  

That said, I see an issue with this approach (even though it will work).  The issue is that this only works if the database already exists (which is does).  If it doesn't, the backend wouldn't be able to start without the database.  

Also, I should probably do the same for test.  I'll make that fix.

afwilcox: I don't see the FLYWAY_LOCATIONS parameter in the backend deploy?  Although it will probably still work as it is the default (I think)

`  deploys-prod-backend:
    name: Deploy Backend to PROD
    needs:
      - deploys-prod-nats
    environment: prod
    runs-on: ubuntu-22.04
    steps:
      - uses: bcgov-nr/action-deployer-openshift@v2.3.0
        with:
          file: backend/openshift.deploy.yml
          oc_namespace: ${{ vars.OC_NAMESPACE }}
          oc_server: ${{ vars.OC_SERVER }}
          oc_token: ${{ secrets.OC_TOKEN }}
          overwrite: true
          parameters: -p ZONE=prod -p PROMOTE=${{ github.repository }}/backend:test
            -p NAME=${{ github.event.repository.name }} -p PROMOTE_MIGRATION=${{ github.repository }}/database-migrations:test
          penetration_test: false`

afwilcox: Okay we'll see how it goes, the test environment appears to be working correctly, and we can re-verify prod.

afwilcox (approved):

PR #463, opened 2024-06-14
feat: CE-782: Create 'Landing page' for users following an invite link

marqueone-ps:
# Description
Create a new view for processing COMS access invites

Story # CE-782

# How Has This Been Tested?

Currently manual testing, not certain this can be automated

- [x] Manual Testing

## Checklist

## Further comments

When testing `StrictMode` gets in the way and when testing a new invite token if the request is made successfully, the request will be processed a second time resulting in a failure. Each time the tests are run it will require a new token which is another problem in and of its self

---

Thanks for the PR!

[Backend](https://nr-compliance-enforcement-463-frontend.apps.silver.devops.gov.bc.ca/api/) available
[Frontend](https://nr-compliance-enforcement-463-frontend.apps.silver.devops.gov.bc.ca/) available


frontend/src/app/components/containers/pages/verify-access.tsx
```diff
@@ -0,0 +1,156 @@
...
+  const [wasRun, setWasRun] = useState(false);
+
+  const environmentName = config.ENVIRONMENT_NAME || "production";
+  const isLoading = useAppSelector((state) => state?.app?.loading?.isLoading);
+
+  useEffect(() => {
+    const token = query.get("token");
+
+    if (!isLoading && !wasRun) {
+      if (token && uuidValidate(token) && !wasRun) {
```
barrfalk, line 34: nitpik: I think the !wasRun is redundant here (since we can only get here if !wasRun is true above).

marqueone-ps: There's some method to that madness. isLoading is waiting for all network requests to finish, and wasRun is tracking the verification. Removing !wasRun will cause an infinite render, and in its current position its causing the infinite render on bad or reused token

barrfalk (approved):

PR #569, opened 2024-07-29
feat: CE-892 expand feature flag

Scarlett-Truong:


backend/src/v1/feature_agency_xref/feature_agency_xref.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+@UseGuards(JwtRoleGuard)
+@ApiTags("feature-agency-xref")
+@Controller({
+  path: "feature-agency-xref",
+  version: "1",
+})
+export class FeatureAgencyXrefController {
+  constructor(private readonly featureAgencyXrefService: FeatureAgencyXrefService) {}
+
+  @Post()
```
marqueone-ps, line 20: you should remove this action if its never used

backend/src/v1/feature_agency_xref/feature_agency_xref.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  @Patch(":id")
+  @Roles(Role.TEMPORARY_TEST_ADMIN)
+  update(@Param("id") id: UUID, @Body() updateAttractantHwcrXrefDto: UpdateFeatureAgencyXrefDto) {
+    return this.featureAgencyXrefService.update(id, updateAttractantHwcrXrefDto);
+  }
+
+  @Delete(":id")
+  remove(@Param("id") id: string) {
+    return this.featureAgencyXrefService.remove(+id);
```
marqueone-ps, line 51: This action should have a Role guard if its being used, if its not being used remove this action

backend/src/v1/feature_agency_xref/feature_agency_xref.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  @Get("features-by-agency/:agencyCode")
+  @Roles(Role.TEMPORARY_TEST_ADMIN, Role.CEEB, Role.COS_OFFICER)
+  async findByAgency(@Param("agencyCode") agencyCode: string) {
+    return await this.featureAgencyXrefService.findByAgency(agencyCode);
+  }
+
+  @Get(":id")
+  findOne(@Param("id") id: string) {
+    return this.featureAgencyXrefService.findOne(+id);
```
marqueone-ps, line 40: This action should have a Role guard if its being used, if its not being used remove this action

frontend/src/app/App.tsx
```diff
@@ -10,26 +10,32 @@ import ColorReference, { MiscReference, SpaceReference } from "./components/refe
...
 import { VerifyAccess } from "./components/containers/pages/verify-access";
+import { FeatureManagement } from "./components/containers/admin/feature-management";
 
 const App: FC = () => {
   const dispatch = useAppDispatch();
 
   useEffect(() => {
     dispatch(getOfficerDefaultZone());
-    dispatch(fetchAllCodeTables());
+    if (Roles.CEEB) {
```
marqueone-ps, line 30: This is just a thought, instead of having separate thunks for different roles/agencies it would be better to have one thunk and have it return the correct code-tables. This way it doesn't matter what your role or agency is, the code-table controller will return whats needed

backend/src/v1/feature_agency_xref/feature_agency_xref.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Role } from "../../enum/role.enum";
+import { JwtRoleGuard } from "../../auth/jwtrole.guard";
+
+@UseGuards(JwtRoleGuard)
+@ApiTags("feature-agency-xref")
+@Controller({
+  path: "feature-agency-xref",
+  version: "1",
+})
+export class FeatureAgencyXrefController {
```
marqueone-ps, line 17: I would recommend renaming this controller to just `FeatureController` instead of `FeatureAgencyXrefController`

frontend/src/app/store/reducers/app.ts
```diff
@@ -245,6 +251,13 @@ export const selectOfficerAgency = (state: RootState): string => {
   return agency;
 };
 
+export const checkFeatureActive =
```
marqueone-ps, line 254: I would suggest renaming this to `isFeatureActive`

marqueone-ps (changes requested):
There's a few changes I think you should make, mostly just removing unused boilerplate code, and renaming a few thing, but otherwise this looks really good @Scarlett-Truong.

frontend/src/app/App.tsx
Scarlett-Truong: @marqueone-ps I'm not quite understand what you mean. You suggest having another thunk function with an agency as a parameter?

backend/src/v1/feature_agency_xref/feature_agency_xref.controller.ts
Scarlett-Truong: @marqueone-ps I was thinking about that but with this current naming, it's easier to know which table it's referring to right of the bat, and the name is consistent with other xref tables like `person-complaint-xref` , `attractant-hwcr-xref` 🤔

marqueone-ps: the `person-complaint-xref` is an exception to some rules, the controller and service logic should actually be in the person controller and service, and the `attractant-hwcr-xref` controller shouldn't exist as the code-table controller actually is what serves up the attractants. At some point there's a lot of clean up we need to do and I've got a feeling we'll probably tackle that when we make the move to emerald.

marqueone-ps (approved):

PR #596, opened 2024-08-20
feat: CE-202-Complaint-Type-doesnt-automatically-reclassify-as-an-Enforcement-Complaint

dmitri-korin-bcps:
# Description

This PR is to set the complaint type default to the type that’s appropriate to the tab/view a user is creating it from.

Fixes # (CE-202)

# How Has This Been Tested?

- [ ] Switch to Enforcement tab of the list/map view. Click ‘Create complaint’ button on the side bar. Check if the complaint type in the create form prepopulates to 'Enforcement'. 
- [ ] Switch to any other tab. Click ‘Create complaint’ button on the side bar. Check if the complaint type in the create form prepopulates to 'Human Wildlife Conflict'. 

## Checklist

### DIFF
--- a/frontend/src/app/components/containers/complaints/complaints.tsx
+++ b/frontend/src/app/components/containers/complaints/complaints.tsx
@@ -1,7 +1,7 @@
-import { FC, useState, useContext, useCallback } from "react";
+import { FC, useState, useContext, useCallback, useEffect } from "react";
 import { shallowEqual } from "react-redux";
 import { Button, Collapse, Offcanvas } from "react-bootstrap";
-import { useAppSelector } from "../../../hooks/hooks";
+import { useAppSelector, useAppDispatch } from "../../../hooks/hooks";
 import { ComplaintFilter } from "./complaint-filter";
 import { ComplaintList } from "./complaint-list";
 
@@ -18,6 +18,7 @@ import { selectCurrentOfficer } from "../../../store/reducers/officer";
 import UserService from "../../../service/user-service";
 import Roles from "../../../types/app/roles";
 import Option from "../../../types/app/option";
+import { setActiveTab } from "../../../store/reducers/app";
 
 type Props = {
   defaultComplaintType: string;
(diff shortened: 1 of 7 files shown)


frontend/src/app/components/containers/complaints/details/complaint-details-create.tsx
```diff
@@ -89,9 +90,15 @@ export const CreateComplaint: FC = () => {
...
-  );
+  let initialComplaintType: string = COMPLAINT_TYPES.HWCR;
+  if (agency === "EPO") {
+    initialComplaintType = COMPLAINT_TYPES.ERS;
+  } else if (agency === "COS") {
+    initialComplaintType = activeTab === COMPLAINT_TYPES.ERS ? COMPLAINT_TYPES.ERS : COMPLAINT_TYPES.HWCR;
+  }
+
+  const [complaintType, setComplaintType] = useState<string>(initialComplaintType);
+
```
marqueone-ps, line 101: @dmitri-korin-bcps you may want to check the users roles rather than using the agency. This would be in line with how you dealt with the filters

frontend/src/app/components/containers/complaints/complaints.tsx
```diff
@@ -18,6 +18,7 @@ import { selectCurrentOfficer } from "../../../store/reducers/officer";
 import UserService from "../../../service/user-service";
 import Roles from "../../../types/app/roles";
 import Option from "../../../types/app/option";
+import { setActiveTab } from "../../../store/reducers/app";
```
marqueone-ps, line 21: simple change here, you've re-imported the same reducer more than once. you'll want to move `setActiveTab `to line 12 where you're you're importing `selectDefaultZone`

marqueone-ps (changes requested):
Just the one change needed for the imports, though I would recommend the change for the roles.

marqueone-ps (approved):

PR #597, opened 2024-08-20
feat: CE 981 hide zone and tabs for CEEB users

gregorylavery:
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-597-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-597-frontend.apps.silver.devops.gov.bc.ca/api)


frontend/src/app/types/app/menu-item.ts
```diff
@@ -3,4 +3,5 @@ export default interface MenuItem {
   name: string;
   icon: string;
   route?: string;
+  roles?: string;
```
marqueone-ps, line 6: This is a great idea, but instead of a string, you'll want this to be an array, this way you can can do a better check for a role

frontend/src/app/components/containers/layout/side-bar.tsx
```diff
@@ -110,7 +115,12 @@ export const SideBar: FC = () => {
       {/* <!-- menu items for the organization --> */}
       <ul className="nav nav-pills flex-column mb-auto comp-nav-item-list">
         {menueItems.map((item, idx) => {
-          return renderSideBarMenuItem(idx, item);
+          if (UserService.hasRole(Roles.CEEB) && !item.roles?.includes('CEEB'))
+            {
+              // Do not display this hence return null
+              return null;
+            }
+            return renderSideBarMenuItem(idx, item);
```
marqueone-ps, line 123: there's a better way to do this, but it does involve making a change to your roles property to an array. Instead of returning a null this way you'll want to do something like this
```
          if (!UserService.hasRole(Roles.CEEB) && !item.roles?.includes("CEEB")) {
            return renderSideBarMenuItem(idx, item);
          }
```

the map function will still return a null, you just don't need to be explicitly returning a null

marqueone-ps (changes requested):
Just a couple things here worth changing. Adding a role property to the menu-item is a great idea

marqueone-ps (approved):

frontend/src/app/types/app/menu-item.ts
gregorylavery: Updated the roles to be a string array. Marking as fixed.

PR #619, opened 2024-09-06
fix: CE 1017

gregorylavery:
Makes the filters dependent upon a feature flag and adds in the DB configurations to set them up for COS, CEEB and PARKS.

# Description

Fixes # (issue)

# How Has This Been Tested?

Logged into the system using both COS and CEEB roles to confirm the listing of the filters is configured correctly for each user type. Looking into updates for Cypress Testing for this now.

- [ ] Test A
- [ ] Test B

## Checklist


afwilcox:
@gregorylavery - Only one thing and this is more of a comment on the feature flag implementation than the work you did...   

the feature_agency_xref table is missing a compound Unique Key constraint on the Feature code and agency_code columns.

This means that these statements in the repeatable script will insert duplicate rows instead of detecting that a row for the species filter for COS already exists:

```
INSERT INTO
  feature_agency_xref (
    feature_code,
    agency_code,
    active_ind,
    create_user_id,
    create_utc_timestamp,
    update_user_id,
    update_utc_timestamp
  )
SELECT
  'SPEC_FLTR',
  'COS',
  'Y',
  user,
  now (),
  user,
  now () ON CONFLICT DO NOTHING;
```

afwilcox (approved):

PR #679, opened 2024-09-27
fix: Ce 1112

gregorylavery:
Uses the agency of the user to filter out a list of Officers for the filtering of complaints.
# Description

Fixes # (issue)

# How Has This Been Tested?

- [ ] Log in as COS officer, verify the listing shows the named officers for that agency.
- [ ] Log in as CEEB officer, verify the listing shows the named officers for that agency.

## Checklist

### DIFF
--- a/frontend/src/app/components/containers/complaints/complaint-filter.tsx
+++ b/frontend/src/app/components/containers/complaints/complaint-filter.tsx
@@ -12,7 +12,7 @@ import {
   selectComplaintStatusWithPendingCodeDropdown,
   selectGirTypeCodeDropdown,
 } from "../../../store/reducers/code-table";
-import { selectOfficersByAgency, selectOfficersByAgencyDropdown, selectOfficersDropdown } from "../../../store/reducers/officer";
+import { selectOfficersByAgencyDropdown, selectOfficersDropdown } from "../../../store/reducers/officer";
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/containers/complaints/complaint-filter.tsx
```diff
@@ -49,6 +49,7 @@ export const ComplaintFilter: FC<Props> = ({ type }) => {
     officers.unshift({ value: "Unassigned", label: "Unassigned" });
   }
   const agency = getUserAgency();
+  const officersByAgency = useAppSelector(selectOfficersByAgencyDropdown(agency));
```
afwilcox, line 52: Almost... You dropped some of the logic off the officers object that puts Unassigned at the top of the list (this is why Cypress tests are important)

You probably could just modify line 47 to use this selector and then you don't need this line, or the change you made below.

afwilcox (changes requested):

frontend/src/app/components/containers/complaints/complaint-filter.tsx
gregorylavery: For some reason I'd thought this officers listing was used elsewhere so needed a separate instance. I've since renamed it as officersByAgency to indicate what it holds and combined the 'Unassigned' code with that.

afwilcox (approved):

PR #713, opened 2024-10-17
fix: CE-909 Animal outcome drug forms reorder unexpectedly

joshgamache:
Addresses [CE-909](https://env-ceds.atlassian.net/browse/CE-909). Addresses issue where data entered into one animal outcome drug field would swap around to other drugs. 

## Changes 🚧

- Drug items now use an ordering function to ensure consistent order, and therefore entries. 

## How Has This Been Tested? 🔬

- [x] Manual testing, matching replication procedure from issue report.

### To test 🔬

1. Run a local instance of the application, log in and create a new issue. 
2. Under _'Outcome by Animal'_, click _'Add animal'_.
3. In animal outcome form, click 'Add drug' multiple times.
4. In the first record, enter information in the fields. Information entered should stay in that record. Information entered into other records should stay in their respective record.

### Checklist 📃

[CE-1120]: https://env-ceds.atlassian.net/browse/CE-1120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Addresses #issue_number. brief_description_of_change 

[CE-909]: https://env-ceds.atlassian.net/browse/CE-909?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-713-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-713-frontend.apps.silver.devops.gov.bc.ca/api)

### DIFF
--- a/frontend/src/app/components/containers/complaints/outcomes/oucome-by-animal/create-outcome.tsx
+++ b/frontend/src/app/components/containers/complaints/outcomes/oucome-by-animal/create-outcome.tsx
@@ -307,7 +307,9 @@ export const CreateAnimalOutcome: FC<props> = ({ index, assignedOfficer: officer
     const { drugs: source } = data;
 
     const items = source.filter(({ id }) => id !== drug.id);
-    const update = [...items, drug];
+    const update = from([...items, drug])
+      .orderBy((item) => item.order)
+      .toArray();
 
     updateModel("drugs", update);
   };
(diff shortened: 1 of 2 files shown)


afwilcox (changes requested):
Looks good from a data entry perspective, but it looks like the order is messed up when you press save.

For example if I enter two drugs at once, they are switched around when I press save, however if I enter 1, press save, then add a second they appear in the correct order.

joshgamache:
> Looks good from a data entry perspective, but it looks like the order is messed up when you press save.
>
> For example if I enter two drugs at once, they are switched around when I press save, however if I enter 1, press save, then add a second they appear in the correct order.

@afwilcox Addressed the reordering after save with commit 07222195e1b23e514e63ba27e9b87832c98fad6d

afwilcox (approved):

PR #720, opened 2024-10-22
fix: use BaseComplaint usage from nrs-ce-common-types

gregorylavery:
# Description

Adds a dependency on nrs-ce-common-types as a new package that will include the BaseComplaint type to be shared between the backend and frontend to standardize the objects.

Fixes # (issue)

# How Has This Been Tested?

- [ ] Run the apps using both COS and CEEB roles and confirm that the map and list view work with all the complaint types
- [ ] Open, edit and save complaints by making changes to all fields in the data and confirm they are saved as expected.

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-720-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-720-frontend.apps.silver.devops.gov.bc.ca/api)


afwilcox (changes requested):
New repo should be renamed to nr-compliance-enforcement-common

exported packages could stay the same name for now, but it might be good to rename this to common if the we will not be exporting multiple different packages in the future (can still be nrs-ce-common)

I think the following can be added to the common type:
status (string)
ownedBy (string)
reportedOn (date)
updatedOn (date)
complaintMethodReceivedCode (string)

These fields should be removed from our backend type as they are in the common type:
createdBy 
updatedBy 

These fields should be removed from our frontend type as they are in the common type:
reportedByOther
incidentDateTime

gregorylavery:
@afwilcox 
Updates requested have been addressed - more fields have been moved into the common types structure and the version in use of that package has been updated to v1.0.9 to address this.  The repo name for the common code has also been updated to nr-compliance-enforcement-common

afwilcox (approved):

PR #721, opened 2024-10-23
fix: CE-1043 investigate issue with pipeline image GitHub images appear to be deleted while in use

barrfalk:
# Description

When migrating to production, we shouldn't be rebuilding the release, since this could introduce a change that wasn't tested in the test environment.  Instead, we should be grabbing the latest PR that was merged into the release branch, and use that to deploy that image to production.  I believe this will fix the issue with the deleted images in GHCR, as well as fix the issue where the image tags couldn't be found when deploying to production.

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-721-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-721-frontend.apps.silver.devops.gov.bc.ca/api)


.github/actions/get-latest-pr-number/action.yml
```diff
@@ -0,0 +1,39 @@
...
+  steps:
+    - id: vars
+      shell: bash
+      run: |
+        git fetch origin
+
+        release_branch="${{ github.event.pull_request.head.ref }}"
+        echo "Detected release branch: $release_branch"
+
+        latest_pr=$(git log origin/$release_branch --merges --pretty=format:'%s' | grep -oP '(?<=#)\d+' | head -n 1)
```
jon-funk, line 28: Running this locally and looking through our git log it appears we have fast-forward merges so this returns empty. Removing the `--merges` makes this return, what I think is the correct list, and that does evaluate to the latest PR number. 

At the time of testing, release/noble-sea-lemon had a merge PR #724, and that was returned (`724`) after removing the `--merges` flag

jon-funk (changes requested):
As far as I can reason about the changes and testing what I can locally, I think this will work after removing the `--merges` flag

.github/actions/get-latest-pr-number/action.yml
barrfalk: Nice find!  I'll commit that change.

jon-funk (approved):

PR #724, opened 2024-10-25
fix: CE-1108

gregorylavery:
# Description

Modifies how the filenames are generated for exports when the users are from different agencies.

Fixes # CE-1108

# How Has This Been Tested?

- [ ] As a COS user export a HWCR and confirm that the filename for the PDF generated is in the correct format.
- [ ] As a COS user export a ERS and confirm that the filename for the PDF generated is in the correct format.
- [ ] As a COS user export a GIR and confirm that the filename for the PDF generated is in the correct format.

- [ ] As a CEEB user export a ERS and confirm that the filename for the PDF generated is in the correct format.

## Checklist


afwilcox (changes requested):
Cypress tests will need updates with the new names.

afwilcox (approved):

PR #735, opened 2024-11-04
chore: Fix mapSearch agency check for CEEB users

mikevespi:
# Description

Complaints service's `mapSearch` function uses the local function `_getAgencyByUser` to determine a users agency. This function identifies a users agency through the office they are assigned to. CEEB users have no office assigned, thus `_getAgencyByUser` fails. To get around this, `mapSearch` now uses the `hasCEEBRole` variable already included to determine if `_getAgencyByUser` should be called, or "EPO" should be assigned (as is the case for CEEB users). Current implementation of tests do not address the problem, and there was an existing false positive in the tests for the complaints service. This will be addressed separately.

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-735-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-735-frontend.apps.silver.devops.gov.bc.ca/api)

### DIFF
--- a/backend/src/v1/complaint/complaint.service.ts
+++ b/backend/src/v1/complaint/complaint.service.ts
@@ -129,7 +129,7 @@ export class ComplaintService {
 
     const result = await builder.getOne();
     //-- pull the user's agency from the query results and return the agency code
-    if (result.office_guid && result.office_guid.agency_code) {
+    if (result.office_guid?.agency_code) {
       const {
         office_guid: { agency_code },
       } = result;
--- a/backend/src/v1/complaint/complaint.service.ts
+++ b/backend/src/v1/complaint/complaint.service.ts
@@ -128,12 +128,15 @@ export class ComplaintService {
       .where("officer.user_id = :idir", { idir });
 
     const result = await builder.getOne();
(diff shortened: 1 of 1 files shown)


backend/src/v1/complaint/complaint.service.ts
```diff
@@ -128,12 +128,15 @@ export class ComplaintService {
       .where("officer.user_id = :idir", { idir });
 
     const result = await builder.getOne();
-
     //-- pull the user's agency from the query results and return the agency code
-    const {
-      office_guid: { agency_code },
-    } = result;
-    return agency_code;
+    if (result.office_guid && result.office_guid.agency_code) {
```
afwilcox, line 132: I think this can be shortened as per the sonar cloud warning:

if(result.office_guid?.agency_code){

afwilcox (changes requested):

afwilcox (approved):

PR #753, opened 2024-11-19
feat: QuickClose CE-1178 and CE-1204

nayr974:
https://env-ceds.atlassian.net/browse/CE-1178
https://env-ceds.atlassian.net/browse/CE-1204

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-753-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-753-frontend.apps.silver.devops.gov.bc.ca/api)


backend/src/v1/case_file/case_file.service.ts
```diff
@@ -234,6 +234,26 @@ export class CaseFileService {
     ${caseFileQueryFields}
     }`;
 
+    // If the assessment being updated has no action required or the justification is not duplicate, then check if there
+    // was previously a link to another complaint. If there is, remove it.
+    if (
+      modelAsAny.updateAssessmentInput.assessmentDetails.actionNotRequired ||
+      modelAsAny.updateAssessmentInput.assessmentDetails.actionJustificationCode !== "DUPLICATE"
+    ) {
```
mikevespi, line 242: This works in the intended cases, but it also catches changes to the date or the officer when the link itself isn't changing. We'll need to check what the values are currently against what they're changing to here, so we could potentially pass this function the existing outcome from the frontend to compare against without having to fetch it here.

frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
```diff
@@ -478,17 +479,22 @@ export const HWCRComplaintAssessment: FC<Props> = ({
     if (selectedActionRequired?.value === "No" && !selectedJustification) {
       setJustificationRequiredErrorMessage("Required when Action Required is No");
       return true;
+    } else if (linkedComplaintData?.length > 0 && !linkedComplaintData[0].parent) {
+      setJustificationRequiredErrorMessage(
+        "Other complaints are linked to this complaint. This complaint cannot be closed as a duplicate.",
+      );
     }
+    console.log(linkedComplaintData);
```
mikevespi, line 488: 🥷

frontend/src/app/components/containers/complaints/outcomes/hwcr-complaint-assessment.tsx
```diff
@@ -478,17 +479,22 @@ export const HWCRComplaintAssessment: FC<Props> = ({
     if (selectedActionRequired?.value === "No" && !selectedJustification) {
       setJustificationRequiredErrorMessage("Required when Action Required is No");
       return true;
+    } else if (linkedComplaintData?.length > 0 && !linkedComplaintData[0].parent) {
```
mikevespi, line 482: This is firing as intended for everywhere except the quick close modal from the complaints search list. There it still allows a parent to link as a duplicate and brings its children with it.

mikevespi (changes requested):

backend/src/v1/case_file/case_file.service.ts
mikevespi: It seems to be creating a new link and archiving the old link which shouldn't be a big deal. We end up with the correct information and I don't think this is a common scenario.

mikevespi (approved):

PR #754, opened 2024-11-19
feat: Ce 515 search filter retention

mikevespi:
# Description

For user efficiency and convenience the search query, filters, active tab, display type (list vs map), and column sort order have been added to the persisted cache to prevent them from being reset every time the complaints search page is visited. The actual filters (this does not include search query, sort order, active tab, map vs list) are stored in their own provider, while the entire set of search parameters (filters, query) and display options (sort order, active tab, map or list view) are all stored in the global store under `complaints` -> `complaintSearchParameters`. While this pattern of storing in multiple places is not ideal, the refactoring required to consolidate them touched more code than was comfortable to change before a significant capacity loss. As such, the initialization of the filter provider has been updated to align itself with stored values, and is kept in sync using the same mechanisms that were already in place.

The only parameter that was not persisted from the search page that may be useful for users is the page of the results that they were viewing. When parameterized and persisted, the page initially behaves as intended however amidst the initial renders an extra call of `getComplaints` is being made with a `page: 1` parameter that would need to get tracked down (it wasn't coming from any of the direct calls in the frontend).

Fixes # (issue)

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-754-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-754-frontend.apps.silver.devops.gov.bc.ca/api)


frontend/src/app/types/app/app-state.ts
```diff
@@ -30,4 +30,5 @@ export interface AppState {
 
   featureFlags: Array<FeatureFlagState>;
   activeTab: string;
+  activeComplaintsViewType: "list" | "map";
```
afwilcox, line 33: I think this needs to be defined in a redux migration or users won't get it without a hard cache clear due to redux persist.

afwilcox (changes requested):
This solution seems to have blown away the default filter functionality:
- When a user access the application (or switches tabs) default filters are applied.  

When I first accessed the PR I did not have my default filters as a COS officer (expected: Status = "Open" and Zone = "My Zone").   This is likely why the cypress tests are failing.

There might need to be some discussion with the BA / PO as to how the default filters will work now, e.g. what constitutes a new session?

frontend/src/app/components/containers/complaints/complaint-list.tsx
```diff
@@ -101,26 +102,39 @@ export const ComplaintList: FC<Props> = ({ type, searchQuery }) => {
 
   const totalComplaints = useAppSelector(selectTotalComplaintsByType(type));
   const defaultPageSize = useAppSelector(selectDefaultPageSize);
+  const storedSearchParams = useAppSelector(selectComplaintSearchParameters);
+  const { sortColumn, sortOrder } = storedSearchParams;
+  const freshSearch = Object.keys(storedSearchParams).length === 2 ? true : false;
```
afwilcox, line 107: I think the sonar comment is accurate here - could be simplified to:

const freshSearch = Object.keys(storedSearchParams).length === 2

frontend/src/app/components/containers/complaints/complaints.tsx
```diff
@@ -163,12 +181,20 @@ export const Complaints: FC<Props> = ({ defaultComplaintType }) => {
 export const ComplaintsWrapper: FC<Props> = ({ defaultComplaintType }) => {
   const defaultZone = useAppSelector(selectDefaultZone, shallowEqual);
   const currentOfficer = useAppSelector(selectCurrentOfficer(), shallowEqual);
-  const filters = getFilters(currentOfficer, defaultZone);
+  const storedSearchParams = useAppSelector(selectComplaintSearchParameters);
+  // If the search is fresh, there are only 2 default parameters set. If more than 2 exist,
+  // this is not a fresh search as the search funtion itself sets more filters, even if blank.
+  const freshSearch = Object.keys(storedSearchParams).length === 2 ? true : false;
```
afwilcox, line 187: Same comment as above.

afwilcox (changes requested):

frontend/src/app/components/containers/complaints/complaint-list.tsx
mikevespi: Ahh good catch

frontend/cypress/e2e/complaint-search.v2.cy.ts
```diff
@@ -147,4 +147,47 @@ describe("Complaint Search Functionality", () => {
...
+    cy.get("#complaint-search").click({ force: true });
+    cy.get("#complaint-search").clear().type("wildlife{enter}"); //-- {enter} will perform an enter keypress
+
+    cy.get("#map_toggle_id").click({ force: true });
+    cy.verifyMapMarkerExists(true);
+
+    cy.get("#multi-point-map")
+      .find("div.leaflet-marker-icon")
+      .should(({ length }) => {
+        expect(length).to.eq(2);
```
afwilcox, line 174: this is failing as I think there is only result that would show on the map.

afwilcox (changes requested):

frontend/cypress/e2e/complaint-search.v2.cy.ts
mikevespi: It should be fixed now, I was using local instead of Cypress directly to pick the data, that's my bad

afwilcox (approved):

PR #768, opened 2024-11-22
fix: CE-1150-multiple-action-taken-saved-to-the-same-complaint

dmitri-korin-bcps:
# Description

This PR fixes an issue when the app does not handle well concurrent changes in complaint actions

Fixes # (CE-1150)

# How Has This Been Tested?

- [ ] Open a CEEB complaint that does not have a decision section populated for edit in two browser tabs. Enter decision data and click save. Enter different decision data (select different action) in the second tab. When you click save on the second tab the app displays an error message and recommends to refresh the page. 
       In the complaint list try to filter by the action you entered in the first tab. The app should find your complaint.  
       In the complaint list try to filter by the action you entered in the second tab (that gave you an error). Your complain should not appear in the list.  

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-768-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-768-frontend.apps.silver.devops.gov.bc.ca/api)


afwilcox (changes requested):
I discussed this change with @nayr974  and we both agree that we think that this validation should be moved out of the reducer and into the case management backend 

I'm guessing that the case management backend is probably doing an upsert and you just need to make sure that the action doesn't exist before inserting it (and then you can return a message if it did)

afwilcox (approved):

PR #795, opened 2024-12-04
chore: pipeline health checks during rollout

jon-funk:
# Description

This new scritpt+action runs in tandem with the helm build to QA it at the infra level, highlighting issues that helm may not report back about. Additionally, this script is configured to timeout just before helm and collect information in the event of a failing/timing out helm build, so you get a report on what might be wrong in-PR.

Example of it catching a probe failure in events: https://github.com/bcgov/nr-compliance-enforcement/actions/runs/12170089403/job/33944486709?pr=795

Example of an all green success: https://github.com/bcgov/nr-compliance-enforcement/actions/runs/12171219493/job/33947826812

# How Has This Been Tested?

- [X] Demonstrated success and failure examples

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-795-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-795-frontend.apps.silver.devops.gov.bc.ca/api)

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-795-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-795-frontend.apps.silver.devops.gov.bc.ca/api)


charts/app/templates/backend/templates/deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       labels:
         {{- include "backend.labels" . | nindent 8 }}
     spec:
+      minReadySeconds: 10
```
mishraomp, line 25: do we need to add extra 10 seconds delay for pod to be ready to serve incoming traffic .

jon-funk: We've had some api errors from fresh pods starting up in test when under load. I have another PR that lightens up the probes and scaling rates from the HPA - but the init container timing is tricky with probes in general. My thinking here was just a bit of extra security, although now that you mention it 10s may be too much. Thoughts?

mishraomp: I would recommend we fix the readiness probe, if required, we add startup probes, but once pod is ready, waiting for 10 seconds, sounds not ideal to me

jon-funk: I should mention that on the application level, our health check on the api backend is 'dumb' in that it just responds to the http request, not actually if the db is connected and happy, so the k8s probe's source of truth is unreliable, so I'm compensating for that on the infra level. For now.

jon-funk: removed these - appreciate the feedback @mishraomp

mishraomp: > removed these - appreciate the feedback @mishraomp

Thanks Jon, I would vote for another ticket in the backlog to fix the readiness and liveness probe in the API :)

.github/scripts/rollout_healthcheck.sh
```diff
@@ -0,0 +1,375 @@
...
+    for deployment in $deployment_list; do
+        deployment_status=$(_health_check_deployment $deployment)
+        if [ "$deployment_status" == "false" ]; then
+            succeeded="false"
+            break
+        fi
+    done
+    for pod in $pod_list; do
+        _pod_running=$(_pod_running $pod)
+        if [ "$pod_status" == "false" ]; then
```
afwilcox, line 173: Is this correct?  Is pod_status available in this context when it's deployed as a local variable in the _pod_running method?

Why not just if [ " (_pod_running $pod)" == "false ]

.github/scripts/rollout_healthcheck.sh
```diff
@@ -0,0 +1,375 @@
...
+
+# global flag to indicate if health check passed
+HEALTH_CHECK_PASSED=0
+
+
+set -e # failfast
+trap 'echo "Error occurred at line $LINENO while executing function $FUNCNAME"' ERR
+
+help_str() {
+    echo "Usage: SKIP_AUTH=true LABEL_SELECTOR=\"app.kubernetes.io/instance=nr-compliance-enforcement-PRNUM\" OC_NAMESPACE=c1c7ed-dev .github/scripts/rollout_healthcheck.sh"
```
afwilcox, line 50: Should the script also contain checks for LABEL_SELECTOR and OC_NAMESPACE to validate they are passed in?

.github/scripts/rollout_healthcheck.sh
```diff
@@ -0,0 +1,375 @@
...
+
+# will be set to 1 if a timeout occurs from an unfinished rollout
+# we use this to fetch triage info
+TIMED_OUT=0
+
+# string of commands to run after health check for additional info
+COMMANDS_TO_RUN=""
+
+# global flag to indicate if health check passed
+HEALTH_CHECK_PASSED=0
```
afwilcox, line 43: This seems counter-intuitive.    It looks like this script sets the value "HEALTH_CHECK_PASSED" = 1 (true) whenever a failure condition is encountered.  

Maybe rename HEALTH_CHECK_FAILED?

afwilcox (changes requested):

afwilcox (approved):

PR #801, opened 2024-12-06
chore: CE-443 Automate COMS access requests

mikevespi:
Add a check for COMS access on sign-in and request it if the user does not have it.

# Description

A field `coms_enrolled_ind` was added to the `officer` table to indicate whether a user has permissions to access the COMS bucket. When a user logs in, if the indicator is false a request is made to the COMS API provisioning permissions for the user based on their role. Currently, `READ_ONLY` users are only given "READ" permissions, all other roles are given full CRUD permissions. The COMS endpoint used (PUT to `/bucket` [docs](https://coms.api.gov.bc.ca/api/v1/docs#tag/Bucket/operation/createBucket) ) provisions the requested permissions to the user whose credentials are being used to make the request, with the `secretAccessKey` allowing them to perform actions on the given bucket. This `secretAccessKey` is not user based, and as such the call to COMS is being done in the backend to avoid exposing that secret. The user roles are stripped off of the user object on the request itself to prevent users from being able wrongly assign extra permissions to themselves. In order to access that information a new backend decorator `@User()` was created. The `@Token()` decorator was updated to grab the token out of the headers if the `authentication` property is not present on the request object itself (which it was not for the incoming request). The COMS request requires new secrets that have been added to the chart.

Developers will need to add the following block to their backend .env files, with the appropriate values:
```
OBJECTSTORE_API_URL=
OBJECTSTORE_ACCESS_KEY=
OBJECTSTORE_BUCKET=
OBJECTSTORE_BUCKET_NAME=
OBJECTSTORE_KEY=
OBJECTSTORE_HTTPS_URL=
OBJECTSTORE_SECRET_KEY=
```

Fixes # [(issue)](https://env-ceds.atlassian.net/jira/software/c/projects/CE/boards/4?selectedIssue=CE-443)

# How Has This Been Tested?

Testing was done by creating a new bucket in dev and logging in with the `coms_enrolled_ind` flag set to false, and ensuring that signing into the app resulted in the call to COMS, that the appropriate permissions being provisioned for the user on that bucket, and that the users `coms_enrolled_ind` was set to true. This was repeated with the COS_OFFICER, CEEB and READ_ONLY roles, removing permissions in between tests.
The standard `dev` bucket was not tested on as I did not have MANAGE rights on the bucket and couldn't remove my permissions to re-run tests.

## Checklist

## Further comments

The COMS enrolled indicator value was added to the profile object in the store as requested in the ticket, however the flag in the store is not used for anything at this point. If there is no intent to use the check beyond login, it could likely be removed.

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-801-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-801-frontend.apps.silver.devops.gov.bc.ca/api)


afwilcox (approved):
Looks great - good job!

PR #879, opened 2025-01-17
feat: Complaint details performance improvements

mikevespi:
Improve load time of Complaint Details page.

# Description

The complaint details page has several sections, each its own component. Each of these components was calling the case management API to fetch the case_file details, and setting them in the state. The responsibility of making that CM call was moved to the wrapping component so that it can be made once per page rather than once per section. This was done for the complaint-details-edit and the quick close modal.

Refactored selectComplaintDetails to use createSelector so it's not instantiating a new object every time triggering component effects.
Memoized a couple dynamically created objects.

Fixes # 1332
https://env-ceds.atlassian.net/jira/software/c/projects/CE/boards/4?selectedIssue=CE-1332

# How Has This Been Tested?

- [ ] Adding an assessment to a complaint that does not have a case file in CM yet and verifying that one is created with the correct information. That Component should then become the the read version with the correct information in the UI.
- [ ] Ensured that quick close modals still follow validation on the complaint being actioned in both the list and complaint details pages

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-879-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-879-frontend.apps.silver.devops.gov.bc.ca/api)

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-879-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-879-frontend.apps.silver.devops.gov.bc.ca/api)


frontend/src/app/components/containers/complaints/details/complaint-details-edit.tsx
```diff
@@ -85,6 +86,11 @@ export const ComplaintDetailsEdit: FC = () => {
 
   const { id = "", complaintType = "" } = useParams<ComplaintParams>();
 
+  useEffect(() => {
+    dispatch(getCaseFile(id));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
mikevespi: I dropped it in from another file and didn't catch it. Just added the dependency and removed the disable 👍

frontend/src/app/components/containers/complaints/list-items/complaint-action-items.tsx
```diff
@@ -62,6 +63,8 @@ export const ComplaintActionItems: FC<Props> = ({
 
   const openQuickCloseModal = async () => {
     document.body.click();
+    await dispatch(getComplaintById(complaint_identifier, complaint_type));
```
nayr974, line 66: no need to await a dispatch

frontend/src/app/components/containers/complaints/details/complaint-details-edit.tsx
```diff
@@ -249,7 +249,7 @@ export const ComplaintDetailsEdit: FC = () => {
       return;
     }
     if (hasValidationErrors()) {
-      dispatch(updateComplaintById(complaintUpdate, complaintType));
+      await dispatch(updateComplaintById(complaintUpdate, complaintType));
```
mikevespi, line 252: Need the await here, otherwise getComplaint fires too early and the new values dont populate the complaint details page as it switches back to read only mode.

nayr974: wow egg on face!
 
It goes against my understanding and experience, but it seems like this does work when using "redux-thunk" middleware which is part of redux toolkit: https://www.reddit.com/r/reduxjs/comments/zuuv04/await_dispatch_async_actionredux_tookitcan_anyone/ 
 
If not using thunks, the standard redux pattern of raising downstream actions would be required to get the chained behaviour.

nayr974 (approved):

PR #882, opened 2025-01-20
feat: CE-518-Prevent-user-from-editing-closed-complaints

dk-bcps:
# Description

This PR is to prevent users from editing closed complaints so that they only update open and active complaints

Fixes # (CE-518)

# How Has This Been Tested?

- [ ] Open a closed complaints: (HWC, Enforcement, GIR). Confirm that editing of both the complaint and outcome report is not possible
- [ ] Reopen the complaint. Check if editing of complaint is available again.

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-882-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-882-frontend.apps.silver.devops.gov.bc.ca/api)


frontend/src/app/store/reducers/complaints.ts
```diff
@@ -105,6 +105,10 @@ export const complaintSlice = createSlice({
 
     setComplaint: (state, action) => {
       const { payload: complaint } = action;
+      let currentComplaint: ComplaintDto = complaint as ComplaintDto;
+      if (currentComplaint) {
+        currentComplaint.readOnly = ["CLOSED"].includes(currentComplaint.status);
```
nayr974, line 110: Instead of adding a pseudo-property on the data object itself, could we follow the same pattern as the "isInEdit" flag that controls the edit behaviour and create an equivalent isReadOnly flag? You could dispatch a "setIsReadOnly" action here instead. The overall logic would remain the same but this is consistent with an existing pattern.

jon-funk:
Testing crunchy cleanup fix

nayr974 (approved):

PR #911, opened 2025-01-29
chore: Remove crunchy limits

afwilcox:
# Description

- Remove CPU and Memory Limits from Crunchy values file
- Also cleaned up some limit related values in other files

Fixes # CE-1373

# How Has This Been Tested?

- [ ] It deploys and works

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-911-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-911-frontend.apps.silver.devops.gov.bc.ca/api)

### DIFF
--- a/charts/crunchy/templates/PostgresCluster.yaml
+++ b/charts/crunchy/templates/PostgresCluster.yaml
@@ -49,9 +49,6 @@ spec:
           requests:
             cpu: {{ .Values.crunchy.pgmonitor.exporter.requests.cpu }}
             memory: {{ .Values.crunchy.pgmonitor.exporter.requests.memory }}
-          limits:
-            cpu: {{ .Values.crunchy.pgmonitor.exporter.limits.cpu }}
-            memory: {{ .Values.crunchy.pgmonitor.exporter.limits.memory }}
 
   {{ end }}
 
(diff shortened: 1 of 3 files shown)


charts/crunchy/templates/PostgresCluster.yaml
```diff
@@ -49,9 +49,6 @@ spec:
           requests:
             cpu: {{ .Values.crunchy.pgmonitor.exporter.requests.cpu }}
             memory: {{ .Values.crunchy.pgmonitor.exporter.requests.memory }}
-          limits:
```
jon-funk, line 52: Change to `limits: ~` to null them properly

database/openshift.deploy.yml
```diff
@@ -28,10 +28,6 @@ parameters:
     value: "50m"
   - name: MEMORY_REQUEST
     value: "250Mi"
-  - name: CPU_LIMIT
```
jon-funk, line 31: We'll need to cleanup this folder after we've confirmed nothing is in use (I don't think it is) so don't worry about this file

jon-funk (changes requested):
Update the limits to `limits: ~` to properly null them. Due to helm/kubernetes patching behavior, the existing limit keys are ignored rather than properly nulled/deleted.

You can see this by running `oc get postgrescluster -oyaml -n c1c7ed-dev`

except for the job stanza, where I confirmed manually that the crd properly deleted the limit key after I nulled it, but you'll see the other limits are still there.

charts/crunchy/templates/PostgresCluster.yaml
jon-funk: to clarify: limits only need to be null'd in the templates file. The values file is fine as-is.

jon-funk (approved):
Lookin good!

PR #1060, opened 2025-05-08
feat: CE-1576-BC-Parks-filter-by-area

dk-bcps:
# Description

This PR is to allow filtering complaints by a park area

Fixes # (CE-1576)

# How Has This Been Tested?

- [ ] Log in as Parks user. Check if the complaint list and map can be filtered using the new "Area" filter.

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-1060-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-1060-frontend.apps.silver.devops.gov.bc.ca/api)


backend/src/v1/complaint/complaint.service.ts
```diff
@@ -934,6 +934,18 @@ export class ComplaintService {
...
 
+  private readonly _getComplaintsByParkArea = async (token: string, area: string): Promise<string[]> => {
+    const { data, errors } = await get(token, {
+      query: `{getParksByArea ( parkAreaGuid: "${area}") { parkGuid, name}}`,
+    });
+    if (errors) {
+      this.logger.error("GraphQL errors:", errors);
+      throw new Error("GraphQL errors occurred");
+    }
+    const parkGuids = data.getParksByArea.length > 0 ? data.getParksByArea.map((park) => park.parkGuid) : ["-1"];
```
nayr974, line 945: What's the purpose of returning the "-1" element instead of an empty array?

nayr974 (commented):
Looks ok, but one question.

backend/src/v1/complaint/complaint.service.ts
dk-bcps: “-1” was used just for consistency with other similar queries. We can use an empty array going forward. It will have same effect.

nayr974: I didn't realize other queries used "-1". Seems a bit strange to me, but ok to keep consistency.

PR #1066, opened 2025-05-10
feat: CE-1600 referral emails

mikevespi:
Add email notifications for referrals.

# Description

- Added a new table email_reference that stores email addresses and relates them to agencies &/or zones
- Added Ches module to external apis that is responsible for making the calls to CHES
- Added an email module to the v1 api that is responsible for generating email data to be sent using the Ches module
- Added email template generator for referrals
- Added shared email addresses for COS Parks and CEEB
- Updated the create referral functionality to then send the generated email based on the referred complaint
- Overwrote the shared inbox email addresses with the ceds address
- Feature flagged the above
- Some refactoring where needed along the way

# IMPORTANT
## Still to do on the ticket:
- There is an order of operations issue with the export. AC request the referred_by agency to be the export attached, however the exporter currently uses owned_by for the agency. Avoiding sending the email before referring the complaint but this will require refactoring the export function.
- Verify all zones have a shared inbox

## Testing instructions
In order to manually test this, you will need to go into the database and do the following:
- Set the feature flag to true for `REFEMAIL`
- Update the `email_reference` table setting email_address to your own

Fixes # CE-1600
https://env-ceds.atlassian.net/jira/software/c/projects/CE/boards/4?selectedIssue=CE-1600

# How Has This Been Tested?

Manually setting all email addresses to my own and referring complaints of each type.

- [ ] Test A
- [ ] Test B

## Checklist

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-1066-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-1066-frontend.apps.silver.devops.gov.bc.ca/api)


backend/src/v1/complaint_referral/complaint_referral.service.ts
```diff
@@ -42,6 +48,21 @@ export class ComplaintReferralService {
...
     this._personService.clearAssignedOfficer(createComplaintReferralDto.complaint_identifier);
 
+    // FEATURE FLAG
+    // If both agencies involved have the feature flag active, send the referral email notification
+    const referredByActive = await this._featureFlagService.checkActiveByAgencyAndFeatureCode(
+      createComplaintReferralDto.referred_by_agency_code,
+      "REFEMAIL",
+    );
+    const referrefToActive = await this._featureFlagService.checkActiveByAgencyAndFeatureCode(
+      createComplaintReferralDto.referred_to_agency_code,
```
nayr974, line 58: Does this check need to be for both agencies? Why not just the sender?

backend/src/v1/email/email.service.ts
```diff
@@ -0,0 +1,173 @@
...
+        },
+      ];
+
+      const senderEmailAddress = user.email ?? process.env.CEDS_EMAIL;
+      const supportEmail = process.env.CEDS_EMAIL;
+      const { given_name, family_name } = user;
+      const senderName = `${given_name} ${family_name}`;
+
+      const recipientList = [];
+      const emailReferences = await this._emailReferenceService.findActiveByAgency(referred_to_agency_code);
```
nayr974, line 73: I'm wondering if this would be a little cleaner if findActiveByAgency took an optional parameter of the geo_organization_unit_code? The switch might not be needed at all and there would be a small if (COS) pass complaint.organization.zone else null conditional instead. Not requesting this change, only a thought.

backend/src/v1/complaint_referral/complaint_referral.service.ts
mikevespi: My thought was this would avoid a scenario where a new agency aren't setup for email notifications and the backend starts throwing errors trying to send them anyways. I can pull it for now and if we need the more fine grained control we can add it later.

backend/src/v1/email/email.service.ts
mikevespi: A good one at that, I had started out that way and it's definitely a little cleaner that way at the moment. I went this route because we had discussed that area specific shared inboxes were a real possibility, and in the list of email addresses from COS there were a couple of seemingly complaint-type specific (e.g. a general_investigations_section and aquatic.invasive.species) so I tried to keep it generic

nayr974: It's not a bad thought if we're going to have future agencies enabled for referrals and added. Might it be better to just raise an exception if there's no email address configured? Something wasn't setup right, throw an error so it gets fixed asap instead of suppressing it?

backend/src/v1/complaint_referral/complaint_referral.service.ts
nayr974: It's not a bad thought if we're going to have future agencies enabled for referrals and added. Might it be better to just raise an exception if there's no email address configured? Something wasn't setup right, throw an error so it gets fixed asap instead of suppressing it?

afwilcox (approved):

PR #1076, opened 2025-05-15
fix: Improve assessment cypress tests

nayr974:
Fix a couple bugs in the tests related to assessments, as well as prevent the creation of infinite numbers of assessments by editing instead of creating an assessment if one already exists.


mikevespi (approved):
👍 if we start to encounter more of these irreversible actions we may want to consider a test only endpoint that resets some data. For now this is clean, very nice

PR #1085, opened 2025-05-21
feat: CE-1669 Authorization Column on ERS Complaints

nayr974:
https://env-ceds.atlassian.net/browse/CE-1669

- Adds a Authorization column that should only be visible to CEEB users on the complaint list.

Corresponding CM PR: https://github.com/bcgov/nr-compliance-enforcement-cm/pull/189

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-1085-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-1085-frontend.apps.silver.devops.gov.bc.ca/api)

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-1085-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-1085-frontend.apps.silver.devops.gov.bc.ca/api)


backend/src/v1/complaint/complaint.service.ts
```diff
@@ -1186,6 +1186,30 @@ export class ComplaintService {
...
+          // inject the authorization id onto each complaint
+          items.forEach((item) => {
+            const caseFile = data.getCaseFilesByLeadId.find((file) => file.leadIdentifier === item.id);
+            if (caseFile?.authorization) {
+              item.authorization =
+                caseFile.authorization.type !== "permit"
+                  ? caseFile.authorization.value
+                  : "UA" + caseFile.authorization.value;
+            }
+          });
```
afwilcox, line 1211: This logic looks like it's working backwards... although I'm not sure why since it looks right.    Should this logic be on the case management side though?

For example things that don't have a UA on the details page have them in the list, and vice versa.

afwilcox (changes requested):
Just one functional issue.

backend/src/v1/complaint/complaint.service.ts
nayr974: It is backwards, fixed. I put the logic here because there's similar logic on the getCaseData function already in this service. I think the most appropriate place would be in the front end, but I think it makes more sense here given that's already done in another place.

afwilcox (approved):

PR #1104, opened 2025-05-30
feat: CE-116 Use BC Basemap, add Basemap layers

nayr974:
https://env-ceds.atlassian.net/browse/CE-116


afwilcox (approved):
I made the one minor sonar change, everything else looks good to me

PR #1107, opened 2025-06-03
fix: Remove extra API calls on some operations

afwilcox:
# Description

Fixes # CE-419

This Pull Requests fixes 3 areas where multiple API calls to the backend were occurring when only one was required.   There were 3 areas targetted:
- Assigning an Officer to a Complaint
- Making a Status Change
- Creating a Complaint

Each of these issues was an independent incident where additional dispatches were being made that were not required.   

Complaint Creation:
- This was the simplest case, a dispatch to get the complaint was being made immediately followed by a navigation to the complaint details page... which also does a dispatch.  The first dispatch was removed.

Officer Changes:
- This was the most complicated case.   Because assignment can occur from both the complaint details page and the list view an extra dispatch was being added to support the list view case.   This PR adds some logic to thunk and an extra parameter to the modal to indicate which scenario is being used.

Status Changes:
- This is a holdover from when you used to be able to update the status from the list view.   Now that you can't any more, rather than parameterizing it in the same way as the Officer modal, this PR just updates the logic to not required the extra dispatch which would have been required to update the list view.

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

## Checklist

- [X] New and existing unit tests pass locally with my changes

## Further comments

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-1107-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-1107-frontend.apps.silver.devops.gov.bc.ca/api)


nayr974 (approved):
Looks good.

PR #1149, opened 2025-06-25
fix: CE-1756 Sector View not working for CEEB

nayr974:
https://env-ceds.atlassian.net/browse/CE-1756


afwilcox:
Is there any point in going forward with this?  This won't be an issue after PR 1148 goes in

afwilcox (approved):

PR #1156, opened 2025-07-03
chore: CE-1713 Terraform Alerts

nayr974:
adds terraform alert definitions for new sysdig namespace


jon-funk (approved):
Yay emerald alerts!

PR #1186, opened 2025-08-06
feat: CE-1667 Create Case

nayr974:
https://env-ceds.atlassian.net/browse/CE-1667
CM PR: https://github.com/bcgov/nr-compliance-enforcement-cm/pull/228

---

Thanks for the PR!

- [Frontend](https://nr-compliance-enforcement-1186-frontend.apps.silver.devops.gov.bc.ca)
- [Backend](https://nr-compliance-enforcement-1186-frontend.apps.silver.devops.gov.bc.ca/api)


frontend/src/app/components/containers/cases/map/case-map.tsx
```diff
@@ -0,0 +1,66 @@
+import { FC } from "react";
```
nayr974, line 1: No change, just moved into a folder.

frontend/src/app/components/containers/cases/list/case-list.tsx
```diff
@@ -0,0 +1,138 @@
+import { FC, useCallback } from "react";
```
nayr974, line 1: No change, just moved into a folder.

frontend/src/app/components/containers/cases/list/case-list-item.tsx
```diff
@@ -0,0 +1,87 @@
+import { FC } from "react";
```
nayr974, line 1: No change, just moved into a folder.

frontend/src/app/components/containers/cases/list/case-filter-bar.tsx
```diff
@@ -0,0 +1,128 @@
+import { FC, MouseEventHandler, useCallback } from "react";
```
nayr974, line 1: No change, just moved into a folder.

frontend/src/app/App.tsx
```diff
@@ -1,11 +1,13 @@
 import { FC, useEffect } from "react";
 import { Route, BrowserRouter as Router, Routes, useParams } from "react-router-dom";
+import { ToastContainer } from "react-toastify";
```
nayr974, line 3: Moved ToastContainer to a central place in the app so it no longer needs to be added to each routes component.

frontend/src/app/components/containers/layout/side-bar.tsx
```diff
@@ -14,7 +14,7 @@ export const SideBar: FC = () => {
   const dispatch = useAppDispatch();
   const isOpen = useAppSelector(isSidebarOpen);
 
-  const menueItems: Array<MenuItem> = [
+  const menuItems: Array<MenuItem> = [
```
afwilcox, line 17: thank youe.

frontend/src/app/graphql/hooks/useGraphQLMutation.ts
```diff
@@ -0,0 +1,40 @@
+import { useMutation, UseMutationOptions, UseMutationResult, useQueryClient } from "@tanstack/react-query";
+import { useRequest } from "@graphql/client";
+
+interface QueryInvalidationOptions {
+  /** Query keys to invalidate on successful mutation */
+  invalidateQueries?: Array<string | Array<any>>;
+}
```
afwilcox, line 7: A discussion point for the future... how do we know which queries to invalidate?

frontend/src/app/App.tsx
afwilcox: nice!

afwilcox (changes requested):
Just making note of missing functionality based on the acceptance criteria.   if this was all discussed and the ticket is out of date let me know:
- There is no officer assigned field when creating the case
- case description is supposed to be mandatory (even though we aren't storing it) which I think is ok... unless we should be?
- The lead agency is supposed to be based on the current user agency instead of selecting it.  Do we want to allow for people to create cases that belong to other agencies?

This one isn't in the AC, and I'm good to merge without doing it, but I noticed when I create a case and then go back to the list of cases, the newly created case doesn't show up.   This isn't the way Natcom works.

frontend/src/app/components/common/form-field.tsx
```diff
@@ -0,0 +1,31 @@
...
+  required?: boolean;
+  validators?: any;
+  render: (field: any) => ReactNode;
+}
+
+/**
+ * Wrapper component that works around a sonarqube rule disallowing setting the
+ * children prop, which is the recommended pattern for tanstack forms.
+ */
+export const FormField: FC<FormFieldProps> = ({ form, name, label, required = false, validators, render }) => (
```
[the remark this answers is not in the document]
nayr974: If we pull the error state rendering to the FormField level and make sure the components interact with standard input field interactions, maybe we can? I will take a future stab at it

frontend/src/app/components/containers/layout/side-bar.tsx
nayr974: welcomue

nayr974:
> Just making note of missing functionality based on the acceptance criteria. if this was all discussed and the ticket is out of date let me know:
> 
>     * There is no officer assigned field when creating the case
> 
>     * case description is supposed to be mandatory (even though we aren't storing it) which I think is ok... unless we should be?
> 
>     * The lead agency is supposed to be based on the current user agency instead of selecting it.  Do we want to allow for people to create cases that belong to other agencies?
> 
> 
> This one isn't in the AC, and I'm good to merge without doing it, but I noticed when I create a case and then go back to the list of cases, the newly created case doesn't show up. This isn't the way Natcom works.

I addressed everything except the officer A/C. We removed the officer selection out of scope of the prior tickets that just didn't get removed out of the written scope of the following tickets that were all written together originally.

I think we should save description, but it is OOS of the intention of this ticket which I see "allow Creating Cases with whatever properties we implemented during the View Cases ticket". I can do it in a standalone PR maybe, given the timing.

afwilcox (approved):

PR #1189, opened 2025-08-07
feat: CE-1812 View complaints that are linked to a case file

dk-bcps:
This PR is to update the case details page to display complaints associated with this case. Also it is to add columns for other features related to a case like: Inspections, Investigations, and Records.

# Description

Fixes # (CE-1812)

# How Has This Been Tested?

**This PR is only to display complaints that are already linked to a case.** Until the feature to link complaints and cases is implemented they have to be linked manually. 

1. [ ] Create a case
2. [ ] Note your case GUID
3. [ ] Connect to the shared database
4. [ ] Add records to a case_activity table. Use case_activity_guid = <GUID from step 2>, case_activity_type = "COMP" and some existing complaint IDs for the case_activity_identifier_ref column
5. [ ] Click the case with GUID from step 2 on the case list page
6. [ ] Check if the complaints you entered in step 4 are displayed in the Complaints column.

## Checklist

### DIFF
--- a/frontend/src/app/components/containers/cases/details/case-details.tsx
+++ b/frontend/src/app/components/containers/cases/details/case-details.tsx
@@ -1,9 +1,16 @@
-import { FC } from "react";
-import { useParams } from "react-router-dom";
+import { FC, useEffect, useState } from "react";
+import { Link, useParams } from "react-router-dom";
 import { CaseHeader } from "./case-header";
 import { useGraphQLQuery } from "@graphql/hooks";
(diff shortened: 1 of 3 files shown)


nayr974 (changes requested):
I have a couple small requests. I also notice that the Edit case button doesn't float right or match the Complaints detail view:

image

Let's add the "Case details" heading back and make sure the button aligns to the right:

image

nayr974 (approved):

PR #1196, opened 2025-08-12
fix: CE-1845

nayr974:
fixes the suggested officers modal per https://env-ceds.atlassian.net/browse/CE-1845


frontend/src/app/store/reducers/app.ts
```diff
@@ -469,15 +469,13 @@ export const getOfficerDefaultZone = (): AppThunk => async (dispatch) => {
 
       if (response.office_guid !== null) {
         const {
-          office_guid: { cos_geo_org_unit: unit, agency_code: agency },
+          office_guid: { cos_geo_org_unit: unit, agency_code_ref: agency_code_ref },
```
afwilcox, line 472: should this just be

office_guid: { cos_geo_org_unit: unit, agency_code_ref }, now that we no longer have an agency object?

afwilcox (changes requested):
Just a minor fix to adjust sonar warning

PR #1213, opened 2025-08-26
fix: CE-1886

nayr974:
If *all* of the complaints in a search *all* have the same coordinates (exact same lat/lng) it causes an error. This PR fixes the bug described here: https://env-ceds.atlassian.net/browse/CE-1886


jon-funk (approved):
🐛 🔨

PR #1220, opened 2025-09-02
feature: CE-1803 Create Party of interest

dk-bcps:
# Description

This PR is the frontend code to allow a user to add a new party of interest and edit it

Fixes # (CE-1803)

# How Has This Been Tested?

- [ ] Navigate to the party create view at:
- [ ] Select a party type: Person or Company and populate required input fields
- [ ] Click Save
- [ ] Check if party is saved successfully
- [ ] Repeat it for another party type
- [ ] On the party view page click Edit button
- [ ] Check if you can edit the party and save data
- [ ] Check if validation that prevent saving party when not all required fields are populated works.

## Checklist

### DIFF
--- a/frontend/src/app/components/containers/parties/edit/party-edit.tsx
+++ b/frontend/src/app/components/containers/parties/edit/party-edit.tsx
@@ -1,0 +1,340 @@
+import { FC, useCallback, useMemo } from "react";
+import { useNavigate, useParams } from "react-router-dom";
+import { useForm, useStore } from "@tanstack/react-form";
+import { z } from "zod";
+import { gql } from "graphql-request";
+import { PartyEditHeader } from "./party-edit-header";
+import { CompSelect } from "@components/common/comp-select";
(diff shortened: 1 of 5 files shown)


frontend/src/app/components/containers/parties/edit/party-edit.tsx
```diff
@@ -0,0 +1,340 @@
...
+      }
+      business {
+        businessGuid
+        name
+      }
+    }
+  }
+`;
+
+const GET_PARTY_TYPE_CODES = gql`
```
nayr974, line 40: I'm not requesting changes, but I am flagging this for some team conversation about how we want to manage code tables for the new schemas. For Agency Codes, Alec extended the existing code-table service in the NatCom backend to query the shared schema.

But, that was partly done because there is existing code that relies on that. I'm not sure if that's the pattern we want for code tables. We can discuss as a team and refactor if we need to later.

PR #1252, opened 2025-09-22
task: CE-1907 Metabase

nayr974:
Add metabase user permissions and pw var


afwilcox (changes requested):
I think with the introduction of the new variable you'll want to also add a FLYWAY_PLACEHOLDER to the environment section for the migration-complaint job in our docker-compose for people that start the app that way.

PR #1537, opened 2026-04-17
feat: CE-2349 CE2345 Investigation edit should be its own page, Start investigation from list view

nayr974:
https://env-ceds.atlassian.net/browse/CE-2349
https://env-ceds.atlassian.net/browse/CE-2345

---

Thanks for the PR!

- [Frontend](https://f208ae-dev-7-natsuite-frontend.apps.emerald.devops.gov.bc.ca)
- [GitOps App](https://gitops-shared.apps.emerald.devops.gov.bc.ca/applications/f208ae-tools/f208ae-dev-7-natsuite)
- [GitOps E2E Logs](https://gitops-shared.apps.emerald.devops.gov.bc.ca/applications/f208ae-tools/f208ae-dev-7-natsuite?resource=&node=batch/Job/f208ae-dev/f208ae-dev-7-e2e-playwright/0&tab=logs) 
- [E2E Reports](https://f208ae-dev-7-natsuite-e2e.apps.emerald.devops.gov.bc.ca/playwright/)
- [Developer Branch for Infrastructure Changes](https://github.com/bcgov-c/tenant-gitops-f208ae/tree/dev-7-20260417-215538)


backend-cm/src/investigation/investigation/investigation.service.ts
```diff
@@ -239,15 +239,33 @@ export class InvestigationService {
...
 
   async create(input: CreateInvestigationInput): Promise<Investigation> {
+    let caseIdentifier = input.caseIdentifier;
+    const leadAgency = input.leadAgency;
+    const investigationStatus = input.investigationStatus ?? "OPEN";
+
+    // Automatically open a case file if one is not provided
+    if (!caseIdentifier) {
+      const today = new Date().toISOString().slice(0, 10);
+      const caseName = `Investigation ${input.name} (${today})`;
```
afwilcox, line 249: Curious how you landed on this as format for case names - just my opinion but I found having a case named 'Investigation...' seemed a bit odd, also what if we want to expand this in the future.  Also the timestamp doesn't have a lot of value as we already have this data in other columns so it just makes for a long name.

Was there a problem with just using the investigation name for the case name?

migration-investigation/migrations/V1.67.0__add_community_to_investigation.sql
afwilcox, line 1: Don't love the column name - as I think "Community" is a COS specific term - but probably not worth changing.

afwilcox (commented):
Just a few things to chat about, I also don't mind pushing it through - overall looks good.

backend-cm/src/investigation/investigation/investigation.service.ts
nayr974: The A/C listed GUID and there's a catch-22 about the GUID not existing until the investigation is created, which depends on the case being created so there would be a multi-step process to achieve that. 

If I used the Investigation ID only it might collide with an existing Case ID since the FE logic doesn't check that, which seemed somewhat likely with people entering "testestest" or some random stuff. I thought pre-pending "Investigation " and date would mitigate the chance of that in practice and we had already acknowledged this is placeholder so didn't think about it too much beyond that.

migration-investigation/migrations/V1.67.0__add_community_to_investigation.sql
nayr974: updated the backend to geo_organization_unit_code_ref and also added community as a column on the list view and as a filter option

afwilcox (changes requested):
Investigation-list.spec.ts will need to be updated with the new column.

afwilcox (approved):

PR #1593, opened 2026-05-14
feat: CE-2403

nayr974:
https://env-ceds.atlassian.net/browse/CE-2403

---

Thanks for the PR!

- [Frontend](https://f208ae-dev-3-natsuite-frontend.apps.emerald.devops.gov.bc.ca)
- [GitOps App](https://gitops-shared.apps.emerald.devops.gov.bc.ca/applications/f208ae-tools/f208ae-dev-3-natsuite)
- [GitOps E2E Logs](https://gitops-shared.apps.emerald.devops.gov.bc.ca/applications/f208ae-tools/f208ae-dev-3-natsuite?resource=&node=batch/Job/f208ae-dev/f208ae-dev-3-e2e-playwright/0&tab=logs) 
- [E2E Reports](https://f208ae-dev-3-natsuite-e2e.apps.emerald.devops.gov.bc.ca/playwright/)
- [Developer Branch for Infrastructure Changes](https://github.com/bcgov-c/tenant-gitops-f208ae/tree/dev-3-20260514-021022)


backend-cm/src/investigation/activity_note/activity_note.service.ts
```diff
@@ -24,7 +26,7 @@ export class ActivityNoteService {
       },
     });
     if (!prismaActivityNote) {
-      throw new Error(`Activity note with guid ${activityNoteGuid} not found`);
+      throw new UnauthorizedAccessException("You do not have access to this activity note.");
     }
```
afwilcox, line 28: I'm not sure this is a good idea - as it confirms the existence of an activity note (although it could also return this error when bad input was put in)

I think usually the pattern is just to leave it at not found.   This might be a broader comment into this whole PR.

edit: we talked about this and want to consider this as a future enhancement to this functionality.

backend-cm/src/pg-session-extension/agency-from-roles.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!clientRoles) return DEFAULT_AGENCY;
+  const roles = Array.isArray(clientRoles)
+    ? clientRoles
+    : clientRoles
+        .split(",")
+        .map((r) => r.trim())
+        .filter(Boolean);
+
+  let agency = DEFAULT_AGENCY;
+  for (const role of ROLE_PRIORITY) {
```
afwilcox: Not sure this makes sense.    Although we can't enforce mutual exclusion via Keycloak a case where a user has multiple roles shouldn't happen.   

The order you've put here is arbitrary and could lead to information leak if for example the user had the COS role by accident.   

If anything I think if a user has multiple agency roles we should fall back to no agency and just let things blow up.

Edit: could also do the same for the case where a user has 0 roles.   Might be nice to have a specific error page saying that the user isn't setup correctly.

migration-investigation/migrations/policies/R__investigation_access_policy.sql
afwilcox, line 1: Do we need to put RLS on all the tables or just the ones that can be accessed via URL e.g. Investigation and Tasks

afwilcox (changes requested):

backend-cm/src/investigation/activity_note/activity_note.service.ts
nayr974: Reverted to the previous behavior

backend-cm/src/pg-session-extension/agency-from-roles.ts
nayr974: Removed this logic and instead throws an error when no role or multiple agency roles

migration-investigation/migrations/policies/R__investigation_access_policy.sql
nayr974: Trimmed this down

afwilcox (approved):

PR #1681, opened 2026-07-30
feat: CE-2594 Add/edit a party when editing a contravention

dk-bcps:
# Description

This PR is to allow changing a party after contravention is created

Fixes # (CE-2594)

# How Has This Been Tested?

- [ ] Add two parties to an investigation. Add a contravention and assign it to one of the parties. Click the Edit action button for that contravention. Check that you can assign another party to this contravention.

## Checklist

---

Thanks for the PR!

- [Frontend](https://f208ae-dev-1-natsuite-frontend.apps.emerald.devops.gov.bc.ca)
- [GitOps App](https://gitops-shared.apps.emerald.devops.gov.bc.ca/applications/f208ae-tools/f208ae-dev-1-natsuite)
- [GitOps E2E Logs](https://gitops-shared.apps.emerald.devops.gov.bc.ca/applications/f208ae-tools/f208ae-dev-1-natsuite?resource=&node=batch/Job/f208ae-dev/f208ae-dev-1-e2e-playwright/0&tab=logs) 
- [E2E Reports](https://f208ae-dev-1-natsuite-e2e.apps.emerald.devops.gov.bc.ca/playwright/)
- [Developer Branch for Infrastructure Changes](https://github.com/bcgov-c/tenant-gitops-f208ae/tree/dev-1-20260730-190949)


backend-cm/src/investigation/contravention/contravention.service.ts
```diff
@@ -197,6 +185,15 @@ export class ContraventionService {
               update_utc_timestamp: new Date(),
             },
           });
+
+          await db.contravention_party_xref.create({
```
nayr974: I don't think we should always create a new xref, we should update the relationship to the new party.

nayr974 (changes requested):
A couple issues:

1. Enforcement actions are lost whenever a save happens, including when no party change is made. It looks like a new xref is being created on each save, I think we should update the existing xref to the new party if changed.

2. When editing a contravention with an unknown party, changes are not saved unless a party is selected.

nayr974 (changes requested):
This fixed the two issues, but introduces a new one. When there is no party (unknown party) now no xref is created, so the contravention does not get assigned to the newly selected party. I think we need some conditional behaviour between the update you have now and the create you had before for this case.

nayr974 (approved):

